PR #29, opened 2015-12-08
Whois update related contacts

teadur:


kritik:
@vohmar I would suggest syntax
```
 related_domain_descriptions.each{ |x, y| WhoisRecord.find_by(name: x).save}	
```

PR #421, opened 2017-03-12
Remove unneeded specs

artur-beljajev:
Merged only with staging


teadur:
There is no min password lenght check anymore or why was the test deprecated ?

artur-beljajev:
I accidentally duplicated this spec when merging, if you mean #min_password_length.

Issue #460, opened 2017-04-10
Simplify seeds.rb


artur-beljajev:
Part of #388

vohmar:
Please use .tld as an example tld instead of .com where ever you need some example domain for tests and seeds - in this case test.tld for the registrar and admin user email. And use country code EE where you need some country code. EE is as good as any other country code so why use US?

artur-beljajev:
Domains/zones have been fixed to use .tld. Country code of US is used for the following reasons:
1. US is more recognisable than EE. I followed the principle that initial data should be as clear as possible
2. Using EE as a country code of admin user requires to specify "identity_code". The point was to keep as less as possible data in seeds.rb. Also, I don't think it's good idea to force any other outside-of-estonia based registry digging in estonia id-code generation.

vohmar:
apparently already in production

PR #478, opened 2017-04-23
Simplify seeds.rb

artur-beljajev:
#460


artur-beljajev:
I just realized, that zone is not needed. It can be created in UI.

Issue #503, opened 2017-05-04
Improve db setup


artur-beljajev:
Remove `db/schema-read-only.rb` and keep using `db/structure.sql` only. Merging the former requires additional effort with no benefit.

Simplify lib/tasks/db.rake and use standard rake tasks as much as possible.

Remove rake tasks:
- db:all:setup
- db:all:create
- db:all:drop
- db:all:schema:load
- db:all:schema:dump

Introduce new rake tasks (custom):
- db:setup:all
- db:schema:load:all
- db:schema:dump:all

artur-beljajev:
@ratM1n @teadur FYI

teadur:
Which standard rake tasks provide the functionality of the removed rake tasks ?
If we change/remove tasks we need to update docs.

artur-beljajev:
1. Everything found by running `rake -T | grep db`
2. Docs are updated

ratM1n:
output from all new tasks:

`
HEAD is now at 27ee3b2... Improve db setup
import@staging:~/registry$ RAILS_ENV=staging rake --trace db:setup:all
** Invoke db:setup:all (first_time)
** Invoke environment (first_time)
** Execute environment
config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true

** Invoke db:load_config (first_time)
** Execute db:load_config
** Execute db:setup:all
** Invoke db:create:all (first_time)
** Invoke db:load_config
** Execute db:create:all
registry_test already exists
registry_whois_test already exists
registry_api_log_test already exists
** Invoke db:structure:load (first_time)
** Invoke db:load_config
** Execute db:structure:load
** Invoke db:schema:load:all (first_time)
** Invoke environment
** Invoke db:load_config
** Execute db:schema:load:all

------------------------ api_log_staging schema loading -----------------------------
-- enable_extension("plpgsql")
   -> 0.0347s
-- create_table("epp_logs", {:force=>:cascade})
   -> 0.0271s
-- add_index("epp_logs", ["uuid"], {:name=>"epp_logs_uuid", :using=>:btree})
   -> 0.0083s
-- create_table("repp_logs", {:force=>:cascade})
   -> 0.0120s
-- add_index("repp_logs", ["uuid"], {:name=>"repp_logs_uuid", :using=>:btree})
   -> 0.0084s
-- initialize_schema_migrations_table()
[...]

PR #504, opened 2017-05-04
Improve db setup

artur-beljajev:
#503


artur-beljajev:
Good catch! Deprecation warning is now fixed.

PR #505, opened 2017-05-04
Remove robot bin as unused

artur-beljajev:
@ratM1n @teadur  FYI


artur-beljajev:
Nothing to test. Still got no feedback from @ratM1n or @teadur

vohmar:
@teadur do you confirm?

teadur:
as of we have removed all content beind the bundle exec test (epp tests not rspec tests) we could aswell remove this shellscript what was used by autobot to call out the epp tests.

PR #513, opened 2017-05-09
Test Travis bundler cache

artur-beljajev:


artur-beljajev:
I was unable to benchmark it properly. Let's see how it behaves in real life.

Issue #313, opened 2016-12-27
Registration period and price list management


vohmar:
Starting from 1st of May .ee will have minimum 3 month registrations on one end 

Currently the way to specify supported registration and renew periods is by creating price in the pricelist for the particular action and period. The same goes for subdomains (ie. pri.ee). This means that each time we change the price we have 30 lines in price list to deal with. This needs to be improved.

admin: /admin/zonefile_settings
* the selection in admin settings menu should be renamed to zones
* add new options to the zone settings page
  * list with valid registration periods - text box, contect is validated, expected is comma separated list of rfc compliant periods ie 3m, 6m, 9m, 1y, 2y, 3y, 4y, 5y, 6y, 7y, 8y, 9y, 10y (alternative would be three separate settings min period (3m), max period(10y), step (3m))
  * price differentiation by periods enforced (is price list entry required for each supported period by default the smaller closest price is used and multiplied ie 1y reg price 5€, if registered for 5y the price will be 5 x 5€ = 25€)
  * price differentiation  by action (is separate price list entry required for create and renew actions. By default create is used)
* extra option for subdomain's settings
  * use parent domain's prices and price settings
* price list view (/admin/pricelists) should have validate button with option to specify date - check if all the necessary entries are present now or at specific point in time.
* general logic, if effective price list entry is present use that, if it is missing act as setup requires - if entry must be present return error message, if optional then use closest existing smaller entry and multiply. for remainder look for the next smaller price list entry to add and multiply.

artur-beljajev:
1, 2, 3 y.

artur-beljajev:
def renew
    # Change it when Pricelist model periods change
    max_reg_time = 4.years.from_now

    if renewed_expire_time >= max_reg_time
      add_epp_error('2105', nil, nil, I18n.t('epp.domains.object_is_not_eligible_for_renewal',
                                             max_date: max_reg_time.to_date.to_s(:db)))
    end

artur-beljajev:
#473

vohmar:
@artur-beljajev this statement is incorrect: max_reg_time = 4.years.from_now
* max reg time is currently 3 years and from the 1st of May 10 years
* max allowed expiration date of a domain is **less than** max reg time + 1 (so it is not 11 years it is less than 11 years - it is not allowed to register a domain for 11 years or renew a domain for the next 10 years right after registering it for a year for example.

artur-beljajev:
set default of "days_to_renew_domain_before_expire" to 0

[...]

PR #583, opened 2017-09-01
Refactor settings

artur-beljajev:
Test after #585
Part of #313

Check admin setting page for UI anomalies.


vohmar:
#585 generates errors - so will put thi ticket on hold to wait for the resolution as suggested

artur-beljajev:
Can we push this ticket to prod now?

PR #588, opened 2017-09-16
Set session timeout to 5 years in development and staging environments

artur-beljajev:
- Needed to avoid signing in every 8 hours
- Nothing to test


teadur:
Maybe we should have lower value for production then 8 hours. @ratM1n 
I think something smaller then 5 years for devel/staging would be "nicer".

artur-beljajev:
Georg

> hour for prod, day for staging and 0 for dev

artur-beljajev:
@ratM1n @teadur This is now configurable option
https://github.com/internetee/registry/pull/588/commits/030eaa3078bf8786ae8d13d1f165a63dbba563df#diff-318c31f2f8c28eecdbe0119432a98af0R131

artur-beljajev:
Please make sure all envs are updated with the corresponding values. Session is not limited unless set in application.yml

vohmar:
this is now deployed on staging @teadur please test this configurable feature.

PR #599, opened 2017-09-27
improve registrar portal access denied page

teadur:
use http code (403) when registrar portal access is denied by ip whitelist, show client ip


teadur:
As you can see from the diff it changes the http response code, aswell it shows the client ip you are connecting from, we have had major problems with people not knowing from which ip they cant access the system.

As of i havent written the check_ip method i wont write rspec for it, you can if you want to.
Same goes for the duplication part, i havent written the method as you can see from the commit.
I can change it to erb if you think thats more correct, i went with layout none because thats how it was working before, the page didnot show any layout before, didnot want to change the behviour on my own initative. (rende text doesnot give any layout).

t('access_denied') was used before and is still used, you think we should rather use t('.access_denied') and move the translation to separate file ? You can push to that branch if you want to make it work like that, i dont see the benefit in it myself, as of access_denied will be same for every use case.

artur-beljajev:
I got the point, thanks for clarifying! I will cover it with specs and alter some code, if you don't mind.

teadur:
Sure if you have time for it, i would love to see it get into production asap and wouldnot want to it to get into long issue queue, henche the reason i decided to make it myself rather make it into an issue.

artur-beljajev:
Fixed:  only ipv4 is looked up for non-authenticated user

artur-beljajev:
Please test user switch and IP restriction features in registrar area.

teadur:
When switching to user/role what doesnot have ip whitelisted, user is logoffed and sent to login page without any error.

artur-beljajev:
№ 1: Fixed

teadur:
If user is logged off with reason access is denied from ip, after relogin for first time same user is tried for logging, so the error is shown again, on second retry the user is able to relogin ( correct whitelisted user is picked )

PR #626, opened 2017-10-28
Migrate fabricators to factory bot factories

artur-beljajev:
Production code is not touched, nothing to test


artur-beljajev:
@vohmar It's now ready

PR #630, opened 2017-10-29
Do not load FactoryGirl factories in dev.rake

artur-beljajev:
FactoryGirl is now under "development" group in Gemfile too. No production code changed. Nothing to test.


artur-beljajev:
@vohmar When this could be merged?

PR #683, opened 2018-01-26
Registry 660

artur-beljajev:


artur-beljajev:
If fails because `verify_mode = OpenSSL::SSL::VERIFY_PEER` is used by default, and it seems API host `repp_url` is not trusted.

@teadur @ratM1n Can you take a look?

P.S. When accessing API I use `cert_path` and `key_path` keys in `application.yml`.

Issue #700, opened 2018-02-06
Improve EPP session management


artur-beljajev:
- Add database constraints
- Refactor EPPSession model, EPP::SessionsController, EppController
- Refactor database structure
- Add basic EPP login/logout tests

teadur:
Please consider env['HTTP_COOKIE'] as session source if cookies doesnot give any extra value.

artur-beljajev:
@vohmar @ratM1n Special attention to `db/migrate/20180207071528_extract_user_id_from_epp_sessions_data.rb`

artur-beljajev:
Test:
- EPP login/logout
- EPP session limit

vohmar:
* session limit does not work - successfully created 10 simultaneous sessions under a single registrar

* epp login/logout works
* epp logs in admin seem OK

artur-beljajev:
How can I reproduce № 1?

artur-beljajev:
As far as I remember, we decided to postpone this issue anyway.

teadur:
So the #710 shouldnot be reviewed before #711 is finished ?

artur-beljajev:
@teadur That's up to @vohmar to decide. By "Test EPP session limit" I meant that exactly the same behaviour as in master branch should remain in place. I only did some refactoring concerning that part.

PR #710, opened 2018-02-13
Registry 700

artur-beljajev:


teadur:
cannot review before #711 is fixed as @artur-beljajev mentions in #700

vohmar:
reference to #711 means that session limit will be fixed after that bug is resolved (https://github.com/internetee/registry/issues/729)

Issue #623, opened 2017-10-24
Admin: add vat info to registrar data


vohmar:
For foreign registrars we need to specify whether the registrar is VAT liable.

/admin/registrars/??/edit
New fields to registrar profile are necessary
General block:
* vat %

vat nr field exists in registrar profile - optional

* if address country code is EE then vat% and vat no. values are optional. VAT is registry_vat_rate on the invoice. the value in the vat% field is ignored.
* if address country code is not EE and VAT no. value is missing, vat% value cannot be NULL (range 0-99). VAT is added to the invoice according to the specified vat% value.
* if address country code is not EE and VAT no. value is filled, vat% value is optional. VAT is not applied nor displayed on the invoice.
- Fill line item's "VATCode" with registrar VAT number if present
- Fill invoice's "TotalVAT"

artur-beljajev:
@vohmar It seems VAT country is a required field. Can I use "registrars.country_code" (http://prntscr.com/h9h7hb) to populate existing registrars with VAT country?

artur-beljajev:
See #509

artur-beljajev:
@vohmar `registrars.vat` (a boolean) is not used at all. It has been imported from Fred. Do we need that data? If we do, explain how it will be used.

vohmar:
registrars.vat can be removed

>> It seems VAT country is a required field. Can I use "registrars.country_code" (http://prntscr.com/h9h7hb) to populate existing registrars with VAT country?

lets drop the vat_coutry_code.

artur-beljajev:
`registrars.vat` is removed.

artur-beljajev:
#367 needs to be addressed along with this ticket

vohmar:
if you can solve this in 2 hours the go for it, otherwise we will postpone #367

artur-beljajev:
I messed up the tickets: the one I need is #407. #367 is also important, but can be postponed.

vohmar:
do not see an option to set vat% under a registrar profile in admin (/admin/registrars/52/edit)

artur-beljajev:
This ticket is done in the same branch as #509, which is not ready yet. What's why this ticket is still assigned to me.

artur-beljajev:
e068442409e2908497bf5793ba7ee9403c318d88

artur-beljajev:
Would you like to see registry's VAT rate as the default pre-filled of `VAT rate` form field?

artur-beljajev:
[...]

Issue #754, opened 2018-03-06
Consider normalizing empty values to NULL when inserting into the database


artur-beljajev:
In Rails when saving a model from a form and values are not provided by the user, an empty string is recorded to the database instead of a NULL (mixing blanks and NULLs can become confusing).

virtualfeline:
Do we actually have any case where missing value is acceptable ?
Also consider that NULL returns different results than an empty string in queries. Will this affect smth? Are we sure we should prefer NULL?

artur-beljajev:
@virtualfeline This is the default behaviour, if by missing value you mean empty string.

artur-beljajev:
Example: #746 introduced contact reuse, so if one contact has fax value of "", and another of `NULL`, they will not be considered identical, and therefore will not be reused.
@vohmar  FYI

artur-beljajev:
https://stackoverflow.com/questions/7202319/rails-force-empty-string-to-null-in-the-database

PR #770, opened 2018-03-09
Refactor registrars

artur-beljajev:
Includes
- Under-the-hood changes needed for #623
- UI improvements (default header, breadcrumbs)
- Add tests
- Allow registrar removal if not in use

Test
- Registrars page in admin area


vohmar:
does not deploy

```
       PG::UniqueViolation: ERROR:  could not create unique index "unique_reg_no"
       DETAIL:  Key (reg_no)=() is duplicated.
       :       ALTER TABLE registrars ADD CONSTRAINT unique_name UNIQUE (name);
       ALTER TABLE registrars ADD CONSTRAINT unique_reg_no UNIQUE (reg_no);
       ALTER TABLE registrars ADD CONSTRAINT unique_reference_no UNIQUE (reference_no);
       ALTER TABLE registrars ADD CONSTRAINT unique_code UNIQUE (code);
```

there are empty values in reg_no field in both staging and production dbs

artur-beljajev:
OK, so basically #754 strikes again. Let me think about some quick solution.

artur-beljajev:
I checked it again, and actually #754 has nothing to do with the issue you specified. According to business logic, `registrars.reg_no`, cannot be blank (along with other attributes: https://github.com/internetee/registry/blob/a5869e0ecad1927b47d4e12188cc57508870029e/app/models/registrar.rb#L16)

vohmar:
optional it is then

artur-beljajev:
How about uniqueness check? Do need it?

vohmar:
assuring that if value exists it is unique is nice to have feature

artur-beljajev:
I need your assistance to solve the following problem:

This ticket has a migration which adds missing unique constraint to `registrars.reg_no` (it is required to be present and unique at the application level). The migration fails because of empty values in that column which are obviously not unique. I cannot use `NULL` since that column is declared as `NOT NULL`. So the only way to fix it is to fill with some value. Is it possible?


@virtualfeline, @teadur, @ratM1n

virtualfeline:
Perhaps reg_no should not be unique at all. 

Consider that we have companies from different countries. It is conceivable that business registry numbers might coincide if 2 countries use similar format. 

Also we already had a real case in production when one registrar accredited, left, and came back again as different registrar while having the same company number. So we already do have 2 registrars in production that have the same reg_no value, not only NULLs. 

Also, what is the benefit of having this be unique? Doesn't seem useful in any way.

artur-beljajev:
@virtualfeline No idea about benefit :) I just added a missing constraint (app validated its uniqueness before anyway). I think the problem I described disappears once we drop uniqueness requirement.

If @vohmar confirms it, I will do in in a separate ticket.

vohmar:
blocked by #776

artur-beljajev:
`reg_no` uniqueness validation and database constraint are now removed.

@ratM1n If there is some env where the migration in question has been already applied, you need to manually remove `registrars.unique_reg_no` constraint.

vohmar:
registrar delete returns internal error - registrar with history
`VM104 admin-manifest-5a58585d5fc971be059d49e52316f411f64b18682bd28ea0d4c0c941dd7b7b05.js:4610 POST https://staging-adm.infra.tld.ee:444/admin/registrars/50 500 (Internal Server Error)`

artur-beljajev:
Fixed

PR #781, opened 2018-03-19
Changelog update 180319

vohmar:


artur-beljajev:
Do you want to update staging branch? It is probably easier to just re-create it from master.

PR #815, opened 2018-04-08
Registry 790

### LINKED ISSUE #790: Admin: removing deleteCandidate status must also delete pending delete job
/admin/domains/_id_/edit

it is possible to remove deleteCandidate status from a domain before it is actually deleted., but removing the status does not stop the domain from being deleted the que/cron job remains and is run as intended. removing the status must also remove the job.

---
@ratM1n @teadur `schedule.rb` is updated! CRON task must be updated before deployment.

artur-beljajev:
Closes #790


vohmar:
tests fail!

artur-beljajev:
Fixed

PR #818, opened 2018-04-11
Fix missing translation error in Mail templates page

maciej-szlosarczyk:


artur-beljajev:
https://github.com/internetee/registry/pull/809 should fix Travis build's failure

maciej-szlosarczyk:
`MailTemplate.human_attribute_name :html_body` returns `Html body`, not `HTML body`. In any case, I moved all the templates specific to that controller to its own locale file. `:mail_templates` will live in the main file as it's also used in `base/_menu.haml`.

Issue #840, opened 2018-05-03
EPP: add support for asice format legal documents


vohmar:
Since DigiDOc 3.10 version there are two versions of digitally signed bdoc container formats. One has file extension bdoc (BDOC-TM or ASiC-E LT-TM) and the other has asice extension (BDOC-TS or ASiC-E LT). The LT version has better international compatibility.

We have added only bdoc format to the accepted file format list and have neglected to add asice. We have now a registrar requesting to use it and we should allow it asap.

Need to update XML schemas and probably EPP as well to accept new file format.

vohmar:
getting epp error in the portal for registrara on trying to uplad asice format legaldoc

@staging /admin/epp_logs/52994
```
    <result code="2306">
      <msg lang="en">Attribute is invalid: type</msg>
```

maciej-szlosarczyk:
All of the above are now added to EPP XSD.

PR #842, opened 2018-05-03
Add support for asice document format

maciej-szlosarczyk:


maciej-szlosarczyk:
I guess this should fix #840

PR #854, opened 2018-05-31
Improve CodeClimate config

artur-beljajev:
Nothing to test.

- Show all offences
- Use shared rubocop.yml
- Remove customizations
- Remove unneeded gems


artur-beljajev:
Definitely. Before there was `.rubocop-to-do`, regenerated based on issues at the time of its generation, but I found that this approach just doesn't work. You need to either find&replace file/folder names and/or regenerate it occasionally. It's more effort than benefit. Let's try to minimize the amount of issues with every next change.

artur-beljajev:
What I would do, is to remove this badge from our project front page 

http://prntscr.com/jqkgw5

vohmar:
code climate seems stuck with analysis

maciej-szlosarczyk:
I restarted the thing, the check is complete now.

artur-beljajev:
@vohmar How about removing it (see a comment above)?

vohmar:
@artur-beljajev as we discussed, lets leave the badge as is - removing it does not improve anything except hinding the current state of things from general public. It's an open source project - lets be open.

PR #859, opened 2018-06-05
Remove `lib` folder from autoload path

artur-beljajev:
Test:
- EPP
- EPP query limit (external)
- Country drop-down in UI
- Browse few pages in every portal


app/controllers/epp_controller.rb
```diff
@@ -407,4 +400,26 @@ def session_timeout_reached?
     timeout = 5.minutes
     epp_session.updated_at < (Time.zone.now - timeout)
   end
+
+  def iptables_counter_update
+    return if ENV['iptables_counter_enabled'].blank? && ENV['iptables_counter_enabled'] != 'true'
```
maciej-szlosarczyk: The first part of this condition can be skipped, it's contained within the second one:

```ruby
2.3.7 :004 > ''.blank?
 => true
2.3.7 :005 > '' != 'true'
 => true
2.3.7 :006 > nil != 'true'
 => true
2.3.7 :009 > 'false' != 'true'
 => true
2.3.7 :008 > 'true' != 'true'
 => false
```

app/controllers/epp_controller.rb
```diff
@@ -407,4 +400,26 @@ def session_timeout_reached?
...
+    return if current_user.blank?
```
maciej-szlosarczyk: That's also a bit weird, `current_user` shouldn't be a `String`, it should be an instance of `User` or `Nil`.
`return unless current_user` should communicate the intent better.

artur-beljajev: I don't a see the point in improving code structure. I think the whole feature should be refactored, which requires some effort and is definitely a topic of another branch/ticket.

PR #860, opened 2018-06-05
Improve test env config

artur-beljajev:
@vohmar Test `registrar_ip_whitelist_enabled` and `api_ip_whitelist_enabled` settings (they should remain the same in admin. This ticket turns them both off by default)

@ratM1n or @teadur Make sure the following methods are not used by some CRON jobs:

- `Domain.clean_expired_pendings`
- `Domain.start_redemption_grace_period`
- `Domain.destroy_delete_candidates`

`DomainCron` should be used instead (until the whole thing is refactored).


config/environments/test.rb
```diff
@@ -31,22 +31,18 @@
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
-  ActiveSupport::Deprecation.silenced = true
+  # Randomize the order test cases are executed.
+  config.active_support.test_order = :random
 
-  # For rails-settings-cached conflict
-  config.cache_store = :file_store, 'tmp/cache_test'
+  # Print deprecation notices to the stderr.
```
maciej-szlosarczyk: `Raise an error on deprecation notices`, print to stderr is `:stderr`

artur-beljajev: Fixed

PR #871, opened 2018-06-11
Registry 791

### LINKED ISSUE #791: Admin: cancel force delete should not remove deleteCandidate status
/admin/domains/_id_/edit

cancel force delete function should remove only the statuses that were set on schedule force delete. So deleteCandidate status should not be removed when force delete is canceled even if the status was set due to forceDelete process running to an end. Removing deleteCandidate status must be conscious decision as this is exceptional action.

Furthermore as deleteCandidate status represents domain waiting to be released and not being owned any more by previous registrant cancelling (nor setting for that matter) forceDelete should not be possible until deleteCandidate status is set.

artur-beljajev:
Closes #791


vohmar:
tests fail

artur-beljajev:
Fixed

vohmar:
merge conflicts!

artur-beljajev:
Fixed.

Issue #849, opened 2018-05-22
GDPR - Registrant: Enable user to update contact data of their domain contacts


vohmar:
The goal is to enable users to update their own domain contact data without the need to turn to registrar - each person owns their personal data and should have control over it.

To update the data in the registry db the contact update functionality should be implemented in rest-epp api.

Identically to epp contact update request the user can update only contact data - name, email. phone nr and if postall address processing is enabled postal address as well. User cannot change ident data (ident code, ident type and ident country code)

User can update only their own contact data if a domain has different admin and/or tech contact the user has to turn to registrar to update these details.

To notify registrar about change in domain/contact data we should implement change poll on the epp api (rfc draft: https://tools.ietf.org/html/draft-ietf-regext-change-poll-08).

---
https://github.com/james-f-gould/EPP-Change-Poll

- [x] Registrant API
- [x] EPP Poll
- [x] Registrant area UI

artur-beljajev:
Does EPP API remain unaffected?

artur-beljajev:
#737

vohmar:
Yes Registrant means portal for regsitrant, GDPR there to make it look more important - refering to the policy this functionality was called after.

EPP API will remain unaffected for time being as the portal should use REPP API to provide this.

artur-beljajev:
#239

artur-beljajev:
@vohmar A call to `PATCH /api/v2/contacts/%UUID%` will return contact itself on success. I believe we should revise fields that are returned.

`id` (UUID, #737)
`name`
`email`
`phone`
`fax` (if enabled)
`created_at`
`updated_at`
`ident`
`ident_type`
`ident_country_code`
`auth_info`
`org_name`
`city`
`street`
`zip`
`country_code`
`state`
`statuses`

See also https://github.com/internetee/registry/blob/master/doc/repp/v1/contact.md#response

artur-beljajev:
> User can update only their own contact data if a domain has different admin and/or tech contact the user has to turn to registrar to update these details.

@vohmar Could you elaborate? How it is connected to a domain?

artur-beljajev:
As I see it, it should look like this:

As a Registrant Portal user, who has Estonian ID-card I:
- Open Registrant Portal
- Go to your profile page (similar to one in Registrar Portal)
- Edit my details (say, email)
- Submit the form
- Expect an associated contact (match is done by comparing ID code part from `users.registrant_ident` with `contacts.ident`) to be updated with the new email

[...]

PR #873, opened 2018-06-12
Add UUID db columns

artur-beljajev:
- Needed for REST API v2 (#849)
- Test domain/contact adding/editing, contact/domain transfer
- REST API v1 will probably start returning UUIDs as well, since there is no filter on API side (it just returns all values straight from a DB). Don't think it might be a problem

### DIFF
--- a/app/models/concerns/contact/transferable.rb
+++ b/app/models/concerns/contact/transferable.rb
@@ -16,6 +16,7 @@ module Concerns::Contact::Transferable
     new_contact.regenerate_code
     new_contact.regenerate_auth_info
     new_contact.remove_address unless self.class.address_processing?
+    new_contact.uuid = nil
     new_contact.save(validate: false)
     new_contact
   end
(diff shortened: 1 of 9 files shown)


app/models/concerns/contact/transferable.rb
```diff
@@ -16,6 +16,7 @@ def transfer(new_registrar)
     new_contact.regenerate_code
     new_contact.regenerate_auth_info
     new_contact.remove_address unless self.class.address_processing?
+    new_contact.uuid = nil
```
maciej-szlosarczyk: Why is this required? Is there a reason why UUIDs cannot stay the same after a transfer, like primary keys?

artur-beljajev: Because transfer operation effectively creates new contact, based on the original one.

maciej-szlosarczyk: Folding issues, didn't see the `self.dup` at the beginning. All good 👍 

screen shot 2018-06-13 at 12 29 55

Issue #893, opened 2018-06-22
MID auth does not display code on web interface


vohmar:
MID auth on web interfaces does not display mid auth code in the portal
https://registrar.internet.ee

no code is displayed after entering the phone number - code is sent on the phone though. Entering correct pin results in login.

The sent sms does not identify the environment
Eesti Internet
Authenticating
Kontrollkood
...

maciej-szlosarczyk:
Hey @ratM1n, do we have any messages logged for this error? It would greatly help me debug this thing.

maciej-szlosarczyk:
Issue was caused by problem in the view, we technically have a test for this, but since it's based on Rack::Test, it produced a false positive:

https://github.com/internetee/registry/blob/master/spec/features/registrar/sign_in/mobile_id_spec.rb

The issue should be fixed, but I opened #900 to avoid this kind of problems in the future.

PR #904, opened 2018-07-09
Add Javascript testing base class

### LINKED ISSUE #900: Research and configure javascript testing framework
For certain type of tests, we need a test browser with enabled javascript. For example, #893 would've been caught by a test like that.

## TODO
- [x] Remove Poltergeist (unused, unconfigured)
- [x] Choose appropriate driver (most-likely webdriver)
- [x] Create a new base test class to inherit from
- [x] Configure database_cleaner gem for Minitest
- [x] Update Travis CI configuration
- [x] Update docker config to install required packages inside the docker machine.
- [x] Implement new test for the mobile_id_test

maciej-szlosarczyk:
Fixes #900, inspired by false positive from #893 .
A couple gotchas:

- Since we cannot run these tests in transaction, we need a way to reverse the database after a successful test run. Another option would be to truncate it manually, but that would require us to change how load fixtures. In general we should move away from eager loading all fixtures for all tests, but no need to resolve this issue just yet.
- Puma is also required to run the test, so moved to `:development, :test` gem group.
- Testing depends on Google Chrome and chromedriver. Travis CI is configured properly to run them, new docker images also come with the required software out of the box. @artur-beljajev might need to adjust his dev setup to install required packages. You can use travis configuration and [Dockerfile](https://github.com/internetee/docker-images/blob/master/base_images/ruby/Dockerfile) as a configuration guideline.
- I decided on inheritance, not mixins. So any test classes that need to run javascript and simulate real browser behavior should inherit from `JavascriptIntegrationTest`.
- Poltergeist gem is removed, as it was completely unused and we had no rspec tests written that would use javascript.

---
Can be merged straight to master after it is accepted, since it does not have production impact.


test/integration/registrar/sign_in_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_helper'
+
+class RegistrarSignInTest < JavascriptIntegrationTest
```
artur-beljajev: I would name it "RegistrarAreaSignInTest", mainly to avoid name clash (I have experienced such clashes before, and it was time consuming to eventually find out that `setup/teardown` from 2 classes were merged because of the same name of a test class.

test/test_helper.rb
```diff
@@ -38,4 +40,40 @@ class ActionDispatch::IntegrationTest
     Capybara.reset_sessions!
     Capybara.use_default_driver
   end
-end
\ No newline at end of file
+end
+
+class JavascriptIntegrationTest < ActionDispatch::IntegrationTest
```
artur-beljajev: I think it's worth following Rails 5 naming convention http://guides.rubyonrails.org/testing.html#system-testing, given we anyway migrate it.

So to name it `ApplicationSystemTestCase` instead, and place it into separate file under `test` dir

artur-beljajev:
I wonder if we could avoid using `database_cleaner` gem. I personally don't have any idea :( (besides doing that gem's work manually, which isn't a good one).

maciej-szlosarczyk:
What do you mean run system tests with JS by default? Or rather, what do you mean by system test? I find the Rails nomenclature to be completely arbitrary and just made up because they already had used up the integration test name.

In any case, there's a significant performance penalty (3-5 times slower tests, even more) that I think it's a bad idea to do it do it by default, especially since our application uses very little javascript.

artur-beljajev:
https://en.wikipedia.org/wiki/System_testing
http://guides.rubyonrails.org/testing.html#system-testing

> System tests allow you to test user interactions

Whereas integration tests verify that some parts work together. Seems quite clear for me.

artur-beljajev:
I fully agree regarding speed, but as I mentioned, if it is not run in a real browser with enabled JS by default (using Rack), there is a chance of exactly same false positive, as we had in #893

maciej-szlosarczyk:
>I fully agree regarding speed, but as I mentioned, if it is not run in a real browser with enabled JS by default (using Rack), there is a chance of exactly same false positive, as we had in #893
Any other ideas how to avoid such problem in the future?

We have to swallow it and when see an existing RackTest case that should use headless Chrome we should convert them as we do it. Otherwise we would open another can of worms.

Issue #536, opened 2017-06-01
papertrail


artur-beljajev:

artur-beljajev:
#426

artur-beljajev:
@vohmar Please provide requirement list.

artur-beljajev:
https://docs.google.com/document/d/1GFs_kxVElYoCicZ69vAaUIuol4ri23_WbxJNir0RV6M/edit?ts=59364fb7#heading=h.57eceh79hphg

artur-beljajev:
@vohmar Please provide requirements for 3rd list item in "Requirements" section

vohmar:
domain:
* date + period
* domain name
* contact
  * contact name
  * contact code (personal, business id, birthday)
  * contact country
  * contact type
* status
* host name
* host ip	
* registrar
* epp requests (create, update, renew, transfer, delete)

contact:
* date + period
* contact name
* contact code (personal, business id, birthday)
* contact country
* contact type
* status
* registrar
* epp requests (create, update, renew, transfer, delete) + related epp requests (domain update, domain create)
* domain (domains that contact is associated to)

registrar:
* date + period
* registrar name
* business id
* domain

artur-beljajev:
@ratM1n This needs your input regarding JSON support in Postgres

teadur:
@artur-beljajev Postgres does support indexing on JSON fields.

artur-beljajev:
https://www.postgresql.org/docs/9.4/static/tutorial-inheritance.html

artur-beljajev:
https://wiki.postgresql.org/wiki/Audit_trigger

artur-beljajev:
http://propelorm.org/blog/2011/08/29/introducing-archivable-behavior-and-why-soft-delete-is-deprecated.html

Issue #426, opened 2017-03-24
Upgrade paper_trail


teadur:
We are running paper_trail 4.0.0.beta3(released in May 2015) what is pre-release of papertrail 4.x
https://github.com/internetee/registry/blob/master/Gemfile.lock#L3

We should upgrade to atleast 4.x latest stable what is 4.2.0 (released  2016-05-31)

If we want to move to Rails 5 we need to upgrade to atleast 5.x. if possible upgrading to latest would make sense.

I propose we need to figure out what/if upgrading paper_trails brakes and then upgrade it.
https://github.com/airblade/paper_trail/blob/master/CHANGELOG.md

We can brake the issue to smaller issues and do it in 3 steps.
1) Upgrade to 4.x stable
2) Upgrade to 5.x
3) Upgrade to 6.x

artur-beljajev:
9.2.0

- Every new version of a domain (at least) gets inserted with a duplicate, since `has_paper_trail` if defined twice:
  - One https://github.com/internetee/registry/blob/cefa37d081f39ae73c500067c56d522841ef9db5/app/models/concerns/versions.rb#L7
  - Two https://github.com/internetee/registry/blob/106592e53e8646b9c075ab1e13d5875d332573e4/app/models/domain.rb#L10
- `DomainStatus` does not have corresponding version class. Kind of a bug, introduced by #756. The correct fix is #945
- `undefined local variable or method `paper_trail_version_class' for #<Class:0x00007f8a87ee19c0>` when tried accessing domain history

artur-beljajev:
I wonder how log_*.session column is used.

(edited)
it seems it's always populated by current time

https://github.com/internetee/registry/blob/78d7d5d894999c4b08e10a00ef5a59ee4093fc69/lib/gem_ext/paper_trail.rb#L20

is there any value other than time in prod?
there is created_at column. I guess session can be dropped

PR #914, opened 2018-07-15
Refactor devise integration

artur-beljajev:
Changes:
- DB column `users.password` is renamed to `users.plain_text_password` because of a conflict with Devise
- Users are now scoped
- Translations and config are regenerated
- Global customizations removed

Test:
- Login/logout to all portals, incl. EPP and REPP (incl. Mobile-ID and ID-card)
- Logout from one portal should not log you out from another one
- EPP password change
- Adding/changing new API user


app/controllers/admin/sessions_controller.rb
```diff
@@ -1,28 +1,39 @@
...
+    def after_sign_in_path_for(_resource_or_scope)
+      admin_root_path
+    end
+
+    def after_sign_out_path_for(_resource_or_scope)
+      new_admin_user_session_path
+    end
+
+    def user_for_paper_trail
+      current_admin_user ? current_admin_user.id_role_username : 'anonymous'
```
maciej-szlosarczyk: Another way of expressing the same thing is:
```
current_admin_user&.id_role_username || 'anonymous'
```

app/controllers/registrar/dashboard_controller.rb
```diff
@@ -2,12 +2,6 @@ class Registrar
   class DashboardController < BaseController
     authorize_resource class: false
 
-    def show
-      if can?(:show, :poll)
-        redirect_to registrar_poll_url and return
```
maciej-szlosarczyk: This changes the functionality a bit, are we sure this won't be considered a bug after next release? :)

vohmar:
registrar portal login with username and pw resulted with this message in ui:

> Webserver configuration error: Apache or nginx is missing user name directive. 
> translation missing: en.activemodel.errors.models.depp/user.attributes.base.webserver_missing_client_cert_directive
> No such user 

admin portal login with username and pw resulted in 

> Failed to load resource: the server responded with a status of 403 (Forbidden)

and the message in ui:

> The page you were looking for doesn't exist.

Mobile ID  / id card login works, but in the registrar portal there is this message instead of expected poll message feed welcoming user.:

image

cannot test user create, password change, cert change because of admin auth problems.

artur-beljajev:
> Webserver configuration error: Apache or nginx is missing user name directive
@teadur Was there any Apache config change recently?
@vohmar Show me that error on a screenshot. I don't understand where it comes from

teadur:
@artur-beljajev nope, havent touched apache

vohmar:
invalid password:
image

password is ok and so should be the cert - was working before and is not expired:
image

vohmar:
still getting the exact same error messages. deployed master branch on staging and login with the same cert and parameters was successful.

artur-beljajev:
I have fixed a bug with the following reproducing steps:
1. Fill in a username field with a valid one, and password with the wrong one at `/admin/login`
2. After getting error "Authorization error", type the correct password

Actual behavior:
You get an exception (= "we're sorry" error in production-like envs, where detailed error info is turned off)

Expected behavior:
You're logged in

Is that what you meant by

> admin portal login with username and pw resulted in

?

artur-beljajev:
@maciej-szlosarczyk Do you experience cert problem on your machine?

artur-beljajev:
@maciej-szlosarczyk Do you experience the same issue with certs?

app/controllers/registrar/dashboard_controller.rb
artur-beljajev: Yes, you're right. It alters it a bit. But the problem is that I needed to either test it properly, or abandon that code chunk. Though I wasn't sure that it was worth testing it, since it's not clear whether we need this functionality at all. All that `DashboardController`'s seem like useless. So I guess it's worth creating new ticket and solve that problem separately.

app/controllers/admin/sessions_controller.rb
artur-beljajev: Good point! But if you don't mind, I will leave it as is, since I think it will be changed soon anyway (#426, #536).

artur-beljajev:
> registrar portal login with username and pw resulted with this message in ui:

@vohmar Please retest.

app/controllers/registrar/dashboard_controller.rb
artur-beljajev: Now it's fixed. `DashboardController` stuff is left for another ticket though.

artur-beljajev:
> but in the registrar portal there is this message instead of expected poll message feed welcoming user

Is fixed as well.

vohmar:
* Registrar portal: user certificate is asked on connecting to the portal (before reaching login page) - no good, certificate is user and login method specific. 
  * Site cannot be reached without presenting valid pki cert. PKI cert is not required on id card and mid login.
* admin - after login user is taken to blank page with message login successful. Instead user should be taken to the list of domains.

artur-beljajev:
Fixed.

maciej-szlosarczyk:
Please try to limit changes inside one branch. This is quite difficult to review, test and debug:

image

app/controllers/registrant/sessions_controller.rb
```diff
@@ -97,4 +95,18 @@ def find_user_by_idc(idc)
...
+  def after_sign_in_path_for(_resource_or_scope)
+    registrant_root_path
+  end
+
+  def after_sign_out_path_for(_resource_or_scope)
+    new_registrant_user_session_path
+  end
+
+  def user_for_paper_trail
+    current_registrant_user.present? ? current_registrant_user.id_role_username : 'anonymous'
```
maciej-szlosarczyk: AFAIR, RubyMine for whatever reason suggests usage of ternary operators, but this is a dressed-up [Nil-Check](https://github.com/troessner/reek/blob/master/docs/Nil-Check.md). A more `tell, don't ask` way to express the same thing would be:

```ruby
def user_for_paper_trail
  current_registrant_user&.id_role_username || 'anonymous'
end

File: app/controllers/registrant/domain_update_confirms_controller.rb [hunk 4]
```diff
@@ -19,7 +19,8 @@ def update
                                                           domain_name: @domain.name,
                                                           verification_token: params[:token])
 
-    initiator = current_user ? current_user.username : t(:user_not_authenticated)
+    initiator = current_registrant_user ? current_registrant_user.username :
```
maciej-szlosarczyk: Another null-check.

spec/support/features/session_helpers.rb
```diff
@@ -1,19 +1,19 @@
...
   module SessionHelpers
     def sign_in_to_admin_area(user: create(:admin_user))
-      visit admin_login_url
+      visit new_admin_user_session_url
 
       fill_in 'admin_user[username]', with: user.username
       fill_in 'admin_user[password]', with: user.password
 
-      click_button 'Log in'
+      click_button 'Sign in'
```
maciej-szlosarczyk: This test asserts nothing. If at the end a blank page is rendered, it will pass.

app/controllers/admin/api_users_controller.rb
```diff
@@ -32,7 +32,10 @@ def edit;
     end
 
     def update
-      params[:api_user].delete(:password) if params[:api_user][:password].blank?
+      if params[:api_user][:plain_text_password].blank?
```
maciej-szlosarczyk: This is error-prone. if `[:api_user]` is nil, you get a cryptic error:

`NoMethodError: undefined method '[]' for nil:NilClass`

Preferred solution would be to use [Hash#dig](https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig). It would return `nil` then:

```ruby
2.3.7 :005 > hash = Hash.new
 => {}
2.3.7 :006 > hash[:api_user][:password]
NoMethodError: undefined method `[]' for nil:NilClass
2.3.7 :007 > hash.dig(:api_user, :password)
 => nil
```

app/controllers/registrant/sessions_controller.rb
artur-beljajev: I think `AnonymousUser` class might be even more suitable.

spec/support/features/session_helpers.rb
artur-beljajev: This is not a test, but a test helper (and legacy one, since specs are in maintenance mode).

artur-beljajev:
@maciej-szlosarczyk The rest is old code, I only changed Devise-related things.

app/controllers/registrant/sessions_controller.rb
maciej-szlosarczyk: In general yes, but for that we'd need bigger refactoring, as there are other places that check if `current_user` is NIl or not:

https://github.com/internetee/registry/blob/master/app/controllers/registrant/domain_delete_confirms_controller.rb#L22
https://github.com/internetee/registry/blob/master/app/views/layouts/registrant/application.html.erb#L40

maciej-szlosarczyk:
@artur-beljajev, please fix the `user_for_paper_trail` and let's call it a day.

artur-beljajev:
@maciej-szlosarczyk What is the point fixing it PaperTrail-related stuff, if it's not yet clear, if we will continue using it at all? https://github.com/internetee/registry/issues/536

maciej-szlosarczyk:
That issue was opened a year ago, requires infrastructure changes and is not touched since March.

 I can't imagine it will become a priority within the next 12 months, given that PaperTrail generally works for us and is continuously supported. At least not bigger than auction center and other on-going developments, upgrading Rails version, Ruby version or fixing bugs and improving test coverage.

artur-beljajev:
I would still do a proper refactoring of PaperTrail part in a separate ticket.

maciej-szlosarczyk:
If you say so, let's do that 👍

vohmar:
* problem mentioned in https://github.com/internetee/registry/pull/914#issuecomment-405594384 is still here. No Poll view in registrars' portal. According to epp logs it looks like the poll que is queried only login for the registrar in use, but not on switching the registrar (no record in log)
image

artur-beljajev:
"No Poll view in registrars' portal"

Please provide additional information about the bug. It is not clear whether you see no "polls" menu on first login, or after you switch a user.

artur-beljajev:
I checked it on master, and didn't notice any anomalies. Switching a user leads to the following screen: http://prntscr.com/kjrteg

vohmar:
initial login now takes user to poll message view, but clicking on eid registrar logo takes user to this view: 

/registrar

https://user-images.githubusercontent.com/7734747/44218934-6c9fcd80-a183-11e8-80e7-9c98e5c3adda.png

instead of this:

/registrar/poll

image

artur-beljajev:
Fixed.

PR #915, opened 2018-07-16
Regenerate structure.sql

artur-beljajev:
Starting from version 9.4.17, pg_dump writes schemas explicitly

https://bucardo.org/postgres_all_versions.html#version_9.4.17

This does not directly affect production, since `structure.sql` is only used when a new deployment is set up, so basically there is nothing to test.

@teadur or @ratM1n We use `public` schema on prod, right?


teadur:
@artur-beljajev yes we use public.

PR #921, opened 2018-07-24
Registrant API authentication

### LINKED ISSUE #918: Implement Registrant portal API authentication endpoint
Implement API endpoint that issues tokens based on data from mobile ID and ID card. Should not depend on database authentication, as we currently don't use passwords for such logins. Part of #902 

## TODO

- [x] Find or create users given the identification number and full name.
- [x] Create and encrypt a token
- [x] Implement authentication via bearer token (decrypt the token, check validity, set `current_user`)
- [x] Create base controller and test helpers for both authenticated and unauthenticated scenarios
- [x] Implement ip block on eID authentication
- [x] ~Update module name from `Api` to `API`.~

maciej-szlosarczyk:
Fixes #918


app/controllers/api/v1/registrant/auth_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        before_action :check_ip_whitelist
+
+        rescue_from(ActionController::ParameterMissing) do |parameter_missing_exception|
+          error = {}
+          error[parameter_missing_exception.param] = ['parameter is required']
+          response = { errors: [error] }
+          render json: response, status: :unprocessable_entity
+        end
+
+        def eid
```
artur-beljajev: What does EID stand for? Is that in common use? Wouldn't it be better to use more meaningful name?

config/application-example.yml
```diff
@@ -96,6 +96,8 @@ arireg_host: 'http://demo-ariregxml.rik.ee:81/'
 sk_digi_doc_service_endpoint: 'https://tsp.demo.sk.ee'
 sk_digi_doc_service_name: 'Testimine'
 
+# Registrant API
+registrant_api_auth_allowed_ips: '127.0.0.1,0.0.0.0' #ips, separated with commas
```
artur-beljajev: Given spaces are anyway stripped, I would add it here for better readability.

app/controllers/api/v1/registrant/auth_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        def create_token(user)
+          token_creator = AuthTokenCreator.create_with_defaults(user)
+          hash = token_creator.token_in_hash
+          hash
+        end
+
+        def check_ip_whitelist
+          allowed_ips = ENV['registrant_api_auth_allowed_ips'].to_s.split(',').map(&:strip)
+          return if allowed_ips.include?(request.ip) || Rails.env.development?
```
artur-beljajev: Any chance to remove hardcoded env check and make it configurable?

app/controllers/api/v1/registrant/auth_controller.rb
maciej-szlosarczyk: It's eID, and yes, it's very common:

https://www.ria.ee/ee/eid-info-ja-juhendid.html
https://nortal.com/what-we-do/electronic-identity/

app/controllers/api/v1/registrant/auth_controller.rb
maciej-szlosarczyk: I was considering using an environment variable, but decided against it.

If we use `registrant_api-auth_allowed_ips` (leave blank to allow any IP), that's a bit unclear and leaves a lot of room for configuration error. Figaro does not allow for expressing more complex data structures.

If we create a new variable that's used exactly for allowing all requests to create tokens, that's making it too complicated for ourselves, as the only use of this thing ever is development environment.

It might be worth a try to use a regex to instead of a list, I'd need to investigate and test that.

maciej-szlosarczyk: I did some additional research on this, I thought we could use `ipaddr` library to add some more interesting parsing, such as suport for subnets and IPv6, but it's extremely slow for anything with `/16` or smaller subnet mask. 

/16:
```ruby
Benchmark.bm do |x|
  x.report { ip_addr =IPAddr.new('192.168.1.1/16'); ip_addr.to_range.include?('182.0.0.1') }
  x.report { ip_addr =IPAddr.new('192.168.1.1/16'); ip_addr.to_range.include?('192.168.0.1') }
  x.report { ip_addr =IPAddr.new('192.168.1.1/16'); ip_addr.to_range.include?('0.1.0.1') }
  x.report { ip_addr =IPAddr.new('192.168.1.1/16'); ip_addr.to_range.include?('0.1.2.1') }
end

       user     system      total        real
   0.420000   0.000000   0.420000 (  0.417766)
   0.000000   0.000000   0.000000 (  0.000038)
   0.420000   0.000000   0.420000 (  0.423992)
   0.430000   0.000000   0.430000 (  0.428151)
```

/8:
```ruby
Benchmark.bm do |x|
  x.report { ip_addr =IPAddr.new('192.168.1.1/8'); ip_addr.to_range.include?('0.1.2.1') }
  x.report { ip_addr =IPAddr.new('192.168.1.1/8'); ip_addr.to_range.include?('192.1.2.1') }
end
       user     system      total        real
 100.980000   0.000000 100.980000 (101.159714)
   0.420000   0.000000   0.420000 (  0.419837)
```

The hard-coded environment check is there to stay.

app/controllers/api/v1/registrant/auth_controller.rb
maciej-szlosarczyk: `mobile_id` is a different case, the fact that we use `mid` is just wrong, as there is neither mID nor m-ID, the acronym is made up by us. We should use `mobile_id` or `mobiil_id`. 

I don't think `e_id` is an improvement over `eid`. Most of the usages do not have a dash or an underscore, the only implementation that I was able to find that uses e-ID instead of eID is [Malta](https://www.mita.gov.mt/en/eGov/eGovServices/Pages/e-ID.aspx). Estonia, Belgium, Finland, [Germany](https://www.bsi.bund.de/EN/Topics/ElectrIDDocuments/German-eID/german-eID_node.html) and [European Commission](https://ec.europa.eu/digital-single-market/en/e-identification) all use eID, and generally do not insert dashes or underscores, not even in URLs.

It's the case of `iPhone`, which as a Ruby method should be expressed either as `i_phone` or `iphone`. The first one can be confused with a beginning of a sentence "I phone my doctor get a sick leave", while the second one is most definitely a phone made by Apple corporation. Similar change in meaning happens with MasterCard. Another example of the same problem is `McDonalds`, which becomes either `mc_donalds` (probably a famous rapper, MC Donalds) or `mcdonalds`, a name of a fast-food restaurant. Ruby method names do not allow for conveying capitalisation nuances inside a word without changing it's meaning.

We could use `electronic_id`, but that's an artificial being, it is really uncommon, similarly to `electronic mail`, which is just `email`. Another option is `e_identification` or `e_identity_document`, `e_identity`. Unfortunately, in different countries `id` stands for different things, but the acronym is always common. 

`eid/eID` is the most commonly used version, we are just unfortunate that Ruby cannot represent it properly.

PR #926, opened 2018-07-31
Registrant API domain endpoint

### LINKED ISSUE #922: Registrant API domains endpoint
Implement domain endpoint:

- [x] Get list of all domains that the Registrant is related to (both private and through business registry)
- [x] Get details of a specific domain (find domain by UUID)

Does not include registry lock, should be a separate issue.

maciej-szlosarczyk:
Fixes #922. Unfortunately depends on #918 (as will any other functionality for the API).


config/routes.rb
```diff
@@ -18,6 +18,16 @@
 
   mount Repp::API => '/'
 
+  namespace :api do
+    namespace :v1 do
+      namespace :registrant do
+        post 'auth/eid', to: 'auth#eid'
+
+        resources :domains, only: [:index, :show], param: :uuid
```
artur-beljajev: How about adding `uuid` param constraint?

`constraints uuid: /[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89abAB][0-9a-f]{3}-[0-9a-f]{12}/ do
end`

maciej-szlosarczyk: I don't think there's anything to be gained from that. It technically would save a roundtrip to the database, but it's fairly uncommon to see anything in the mould of `resources :posts, id: /\d+/`, which is far cheaper an easier to maintain. By default, data is requested from the database with even the most bizarre ids. 

screen shot 2018-08-04 at 17 06 34


Not to mention that a constraint that covers the complexity of a UUID, as you commented with is complex enough to be problematic to test.

artur-beljajev: Database round-trip is not relevant, that's for sure. My point was to set up the router to reject invalid requests (say `registrant-api.eis.ee/domains/123456`).

Why `/\d+/`? Don't we use UUID of format x-x-x-x?

maciej-szlosarczyk: > My point was to set up the router to reject invalid requests (say `registrant-api.eis.ee/domains/123456`).

What would be the benefit? The regex is very complex and we both agree that a database query against an index is not that painful.

If there's something to be gained from such a constraint, we should have an analogous one for other resources identified via primary keys, such as `/registrar/invoices/123`, where it would take form of `/\d+/`. We do not.

PR #929, opened 2018-08-01
Change `contacts.name` to NOT NULL

artur-beljajev:
Check prod.


vohmar:
cannot be done as there are empty name values already in production db

artur-beljajev:
If it's absolutely necessary to leave that "broken" contacts there, we could change `NULL` to empty string. If our current policy states that we do require the name, the corresponding column must be declared as `NOT NULL`.

maciej-szlosarczyk:
Alternatively we can also look at the history of these contacts, and see what was the registrar_id in the past, as it should be in Paperclip. And then contact that registrar to make the changes we need.

vohmar:
blocked by https://github.com/internetee/registry/issues/956

artur-beljajev:
#956 is fixed.

vohmar:
blocked until contact archivation updates are in production and we can be sure there are no active contacts with missing name value left

PR #930, opened 2018-08-07
Add contact tests

artur-beljajev:
Nothing to test


test/models/contact_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'test_helper'
+
+class ContactTest < ActiveSupport::TestCase
+  def setup
```
maciej-szlosarczyk: Technically nothing can go wrong here, but You'd probably want to either include `super` or convert this to a block.

test/models/contact_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    @contact = contacts(:john)
+    @original_address_processing_setting = Setting.address_processing
+  end
+
+  def teardown
+    Setting.address_processing = @original_address_processing_setting
```
maciej-szlosarczyk: Technically nothing can go wrong here, but You'd probably want to either include `super` or convert this to a block.

PR #931, opened 2018-08-10
Registry 927 (domain registry lock)

### LINKED ISSUE #927: Implement registry lock for Registrant API
Allow registrant to lock a domain they own or are an person of responsibility for in case of a company domain.

## TODO
- [x] Implement registry lock for domains
- [x] Add functionality to unlock the domains by administrator.
- [x] Expose registry lock via REST API
- [x] Add PaperTrail handling for Registrant API
- [x] Disallow technical contacts from locking a domain.

maciej-szlosarczyk:
Fixes #927


app/controllers/admin/domains/registry_lock_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Admin
+  module Domains
+    class RegistryLockController < BaseController
+
+      def destroy
+        set_domain
+        authorize! :manage, @domain
+        if @domain.remove_registry_lock
+          redirect_to edit_admin_domain_url(@domain), notice: t('admin.domains.registry_lock_delete.success')
```
artur-beljajev: Why not just `notice: t('.deleted')`?

app/controllers/api/v1/registrant/base_controller.rb
```diff
@@ -22,6 +22,15 @@ def bearer_token
           header.gsub(pattern, '') if header&.match(pattern)
         end
 
+        def associated_domains(user)
+          country_code, ident = user.registrant_ident.split('-')
```
artur-beljajev: This part is crying about being encapsulated into `RegistrantUser` :)

app/models/concerns/domain/lockable.rb
```diff
@@ -0,0 +1,53 @@
+module Concerns::Domain::Lockable
+  extend ActiveSupport::Concern
+
+  def apply_registry_lock
```
artur-beljajev: Do we have any other lock besides registry? Simply `#lock` does not work?

app/models/concerns/domain/lockable.rb
```diff
@@ -0,0 +1,53 @@
...
+    lock_statuses = [
+      DomainStatus::SERVER_UPDATE_PROHIBITED,
+      DomainStatus::SERVER_DELETE_PROHIBITED,
+      DomainStatus::SERVER_TRANSFER_PROHIBITED,
+    ]
+
+    (statuses & lock_statuses).count == 3
+  end
+
+  def remove_registry_lock
```
artur-beljajev: I think `apply_registry_lock` is a bad antonym for `remove_registry_lock`. Perhaps `add` would be a better opposite for `remove`?

app/models/concerns/domain/lockable.rb
```diff
@@ -0,0 +1,53 @@
...
+    return unless locked_by_registrant?
+
+    transaction do
+      statuses.delete(DomainStatus::SERVER_UPDATE_PROHIBITED)
+      statuses.delete(DomainStatus::SERVER_DELETE_PROHIBITED)
+      statuses.delete(DomainStatus::SERVER_TRANSFER_PROHIBITED)
+      self.locked_by_registrant_at = nil
+
+      save
```
artur-beljajev: Did you mean `save!`?

app/controllers/admin/domains/registry_lock_controller.rb
maciej-szlosarczyk: Because `Deleted` does not fit in this scenario. It certainly fits less than `Registry lock removed`, as we are not deleting a record, we are removing a restriction.

artur-beljajev: I mean why do you need to spell out the full translation path, instead of using a shortcut?

app/models/concerns/domain/lockable.rb
maciej-szlosarczyk: There's an `ActiveRecord` method called `lock!`, I'd rather avoid such confusion.

app/controllers/admin/domains/registry_lock_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Admin
+  module Domains
+    class RegistryLockController < BaseController
+      def destroy
+        set_domain
```
artur-beljajev: Is it a better way than using `before_action` callback or `.helper_method`?

maciej-szlosarczyk: Yes. there's exactly one method in this controller. `before_action` makes sense when you re-use something more than once.

app/models/concerns/domain/lockable.rb
```diff
@@ -0,0 +1,49 @@
+module Concerns::Domain::Lockable
```
artur-beljajev: Given `ActiveRecord` has `lock!`, wouldn't it be better to name it `RegistryLockable` then?

app/models/concerns/domain/lockable.rb
maciej-szlosarczyk: Maybe `set`, but you rarely `add a lock` or `add a password`, or even `add protection`. You rather `apply` them. All are phrases from similar realm.

artur-beljajev: I like `apply` itself. `apply/remove` pair doesn't look super, though.

app/controllers/api/v1/registrant/base_controller.rb
maciej-szlosarczyk: You're correct, there is even another place where we use a similar block:
https://github.com/internetee/registry/blob/master/app/controllers/registrant/domains_controller.rb#L56-L65

But I'd do it in a separate branch to avoid broad surface area for the manual tests, after this one is in master.

artur-beljajev: I saw some additional chunks that use that `ident` of a `RegistrantUser`.

test/models/domain/domain_version_test.rb
```diff
@@ -0,0 +1,52 @@
...
+      duplicate_domain.save!
+    end
+
+    assert_equal(duplicate_domain.creator, @user)
+    assert_equal(duplicate_domain.updator, @user)
+    assert_equal(duplicate_domain.creator_str, @user.id_role_username)
+    assert_equal(duplicate_domain.updator_str, @user.id_role_username)
+  end
+
+  def test_assigns_updator_to_paper_trail_whodunnit
```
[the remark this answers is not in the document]
maciej-szlosarczyk: I'm testing a very specific implementation, even directly calling the gem internals. I don't think hiding that is beneficial.

maciej-szlosarczyk:
https://github.com/internetee/registry/blame/2dd87d9b065dd43eb0b12d24a75604866f530653/db/structure.sql#L4762

There's an error, but not in that specific migration.

vohmar:
merge conflict

maciej-szlosarczyk:
Fixed now.

PR #935, opened 2018-08-15
Disallow deposits that are lower than 0.01 EUR

### LINKED ISSUE #651: Disallow invoice with zero amount / Balance top-up with zero should fail
Reproducible at least at /admin/invoices/new. Probably applies to registrar area as well.

P.S. Notice the amount of line item.

http://prntscr.com/hdi9v4

maciej-szlosarczyk:
Fixes #651. Also, effectively disallows using a negative `minimum_deposit`. You can still set it to be so, but the minimum invoice in such a case would have to be at least 0.01 €. To test it, try to create deposits with amounts that are not positive.

---
I did it now, because it turns out that the way deposit was written was blocking the upgrade path to Ruby 2.4.


test/models/deposit_test.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def test_validate_amount_cannot_be_lower_than_minimum_deposit
+    @deposit.amount = 0.10
+    refute(@deposit.valid?)
+
+    assert(@deposit.errors.full_messages.include?("Amount is too small. Minimum deposit is 1.0 EUR"))
+  end
+
+  def test_registrar_must_be_set
```
artur-beljajev: How about approach used by Thoughtbot, when you omit association validation in favor of DB constraint, given it's anyway cannot be fixed by a user?

https://robots.thoughtbot.com/validation-database-constraint-or-both

maciej-szlosarczyk: This is not a database table, it's a class that uses ActiveRecord validations. A form object, if you'd like.

artur-beljajev: hm, ok. Missed that fact.

PR #965, opened 2018-09-03
Make whois JSON test order independent

maciej-szlosarczyk:
The test relied on nameservers being inserted in specific order. For
whatever reason, that order can change, but it does not make the
record invalid.

`[ns1.google.com, ns2.google.com]` is semantically equal to
`[ns2.google.com, ns1.google.com]`

### DIFF
--- a/test/models/whois_record_test.rb
+++ b/test/models/whois_record_test.rb
@@ -22,13 +22,16 @@ class WhoisRecordTest < ActiveSupport::TestCase
       registrant_kind: 'priv',
       email: 'john@inbox.test',
       expire: '2010-07-05',
-      nameservers: ['ns1.bestnames.test', 'ns2.bestnames.test'],
       registrar_address: 'Main Street, New York, New York, 12345',
       dnssec_keys: [],
     }
 
     expected_partial_hash.each do |key, value|
       assert_equal(value, @record.generated_json[key])
+    end
+
+    ['ns1.bestnames.test', 'ns2.bestnames.test'].each do |item|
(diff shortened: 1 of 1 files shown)


test/models/whois_record_test.rb
```diff
@@ -22,14 +22,17 @@ def test_generated_json_has_expected_values
...
-      nameservers: ['ns1.bestnames.test', 'ns2.bestnames.test'],
       registrar_address: 'Main Street, New York, New York, 12345',
       dnssec_keys: [],
     }
 
     expected_partial_hash.each do |key, value|
       assert_equal(value, @record.generated_json[key])
     end
+
+    ['ns1.bestnames.test', 'ns2.bestnames.test'].each do |name_server|
```
artur-beljajev: I think it's better to stick to `nameserver`, instead of `name_server` for consistency.

artur-beljajev: https://www.namecheap.com/support/knowledgebase/article.aspx/9434/32/using-default-nameservers-vs-hosting-nameservers

One of the popular domain name sellers also uses single word spelling, whereas WIKI uses separate words.

artur-beljajev: If you want to rename it to `name_server`, let's create a new branch and make it there.

artur-beljajev:
@vohmar Perhaps there should be some specific order in production code?

test/models/whois_record_test.rb
maciej-szlosarczyk: Changed to item, as this is exactly what is important here (specific item being placed in some array of item).

Issue #475, opened 2017-04-17
New registration periods from 1st of May


vohmar:
This is a subticket to https://github.com/internetee/registry/issues/313

This is the absolute minimum functionality that must go live on 1st of May 2017

* epp support for periods 3m, 6m, 9m, 1y, 2y, 3y, 4y, 5y, 6y, 7y, 8y, 9y, 10y (no schema change is necessary)
* portal for registrars support for the periods 3m, 6m, 9m, 1y, 2y, 3y, 4y, 5y, 6y, 7y, 8y, 9y, 10y
* validation update to set the maximum expire date of any domain to be less than maximum allowed period + 1 year (ie expire_time < max_reg_time + 1 year). This rule is necessary to allow domains to be renewed for another max_reg_time (in this case 10 years) period before the domain is actually expired.
* support on the admin interface to enable adding periods to pricelist for the new and old supported periods (new/renew)
* Remove "log_pricelists"

artur-beljajev:
set default of "days_to_renew_domain_before_expire" to 0

artur-beljajev:
https://meedia.internet.ee/files/Domeenireeglid_01.05.2017.pdf

artur-beljajev:
check permissions!

artur-beljajev:
refactor zones

artur-beljajev:
Remove outdated and unused"rake:zonefile:replace_procedure" rake task (lib/tasks/zonefile.rake)"

artur-beljajev:
Timo:

oluline, et epp ja registripidaja portaali kaudu oleks võimalik registreerida ja pikendada domeene 3 kuust 10 aastani

artur-beljajev:
max_reg_time = 4.years.from_now

artur-beljajev:
EPP:create:

sisuliselt on see +1a ja +1p kell 00.00
ehk kui registreerin preagu domeeni timo.ee, praegu on 25.4.17 16.02
siis valid to väärtus peab olema 26.4.17 00.00
eesti aja järgi

artur-beljajev:
Test directo

artur-beljajev:
```
          hash = {
              "ProductID"      => DOMAIN_TO_PRODUCT[pricelist.zone_name],
              "Unit"           => "tk",
              "ProductName"    => ".#{pricelist.zone_name} registreerimine: #{pricelist.years_amount} aasta",
              "UnitPriceWoVAT" => pricelist.price_decimal/pricelist.years_amount
[...]

PR #980, opened 2018-09-16
Return version logging to `prices` DB table

artur-beljajev:
PaperTrail's `Version` model itself has been removed in #475, so those
columns are now useless

### DIFF
--- a/app/models/billing/price.rb
+++ b/app/models/billing/price.rb
@@ -1,6 +1,7 @@
 module Billing
   class Price < ApplicationRecord
     include Concerns::Billing::Price::Expirable
+    include Versions
 
     belongs_to :zone, class_name: 'DNS::Zone', required: true
     has_many :account_activities
--- a/app/models/version/billing/price_version.rb
+++ b/app/models/version/billing/price_version.rb
@@ -1,0 +1,7 @@
+module Billing
+  class PriceVersion < PaperTrail::Version
+    self.table_name    = :log_prices
+    self.sequence_name = :log_prices_id_seq
+  end
+end
+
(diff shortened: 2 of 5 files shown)


vohmar:
merge conflict with master

artur-beljajev:
Changes are not logged. I guess "Update price" button should be taken away http://prntscr.com/l1vmtr. Otherwise I don't see a point of having "Expire" button.

vohmar:
there is a reason for expire - price promotions. The use-case example is

* you have a regular price  beginning from 1.01.2018 without end date
* now you want to have special temporary price promotion - one might add new price list record with starting and end date or if the end date is not set leave end date blank and expire the record once the moment arrives. So the expire button is basically quick action for updating the current price list record's end date to the moment of the button press.
* update price offers more freedom - to change the start date, the price but basically it should be used only if there was some error, In regular operations when one wants to set a new price new record should be created and the last one expired.

It might be worth to discuss if there should be only one button like expire, but either way any changes made to any record incl price list entries should be logged and thus I do not see how could we drop papertrail columns for price list without replacing these with some alternative way of tracking changes.

artur-beljajev:
Just to be sure you understand big picture: currently, there is _no audit log_ for prices at all. Presence of columns in question has no effect on that. They are useless.

Based on your last comment, we need to restore this feature.

vohmar:
the problem is not with the columns but with removed logging / version handling.

artur-beljajev:
@vohmar If I remember correctly, you said it is not needed. Should I close this PR and make a new ticket to restore versioning for `Price` model?

artur-beljajev:
@vohmar Reminder.

vohmar:
Yes we need to track changes with prices as with anything else in the registry - need to restore change tracking. It does not have to be handled by papertrail.

teadur:
possibly will be fixed by #1563

karlerikounapuu (approved):

PR #981, opened 2018-09-16
Remove PaperTrail from historical data models

artur-beljajev:
Removes the following DB tables:
- `log_accounts`
- `log_account_activities`
- `log_invoices`
- `log_invoice_items`

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1,6 +1,4 @@
 class Account < ActiveRecord::Base
-  include Versions
-
   belongs_to :registrar, required: true
   has_many :account_activities
 
--- a/app/models/account_activity.rb
+++ b/app/models/account_activity.rb
@@ -1,7 +1,6 @@
 require 'csv'
 
 class AccountActivity < ActiveRecord::Base
-  include Versions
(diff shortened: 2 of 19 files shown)


vohmar:
This update will not be applied. There is an option to delete a registrar. There is a bug in the system making it currently impossible for registrar accounts with some history, but it works fine for newly created records and in that case log_accounts is the only place record persists.

artur-beljajev:
It has nothing to do with invoices.
However, it is indeed possible to remove a registrar along with his accounts (`accounts` and `white_ips` are the only ones which are removed along with a registrar. I think we should disallow removing any registrar that has at least one associated object (default Rails behavior) and still push this PR.

vohmar:
> How about disallowing to remove any registrar that has at least one associated object (default Rails behavior) and still push this PR?

lets not - first of all I think option to archive inactive registrars is necessary, secondly your proposal still leaves option to "delete" registrars with no associations and as our policy is to keep record of everything. We need history for these records as well. We will come back to historical data management with replacing Papertrail task, lets not remove anything before that.

Issue #329, opened 2017-01-03
Auto-invoicing


vohmar:
to automate credit loading process for registrars it is necessary to add new settings to registrars' profile in the portal for registrars

new options in the portal for registrars. Requires new profile edit view: 
auto-invoicing enabled checkbox
if enabled the 'low credit balance' and 'auto invoice sum' must be set
electronic invoicing option: IBAN (bank account number where to send the e-invoice) - XML
email option: billing email address (registrar can choose one, the other or both options for receiving the invoice) - PDF

if credit balance of the registrar falls to or below the 'low credit balance' setpoint invoice is generated in the sum of 'auto invoice sum' and sent to registrar according to the selected invoice receiving options.

The registrar can now sign auto-payment contract with its bank to fully automate the process.

Registrar users that have access to setting auto-invoicing options must have security profile billing or super (users with epp role cannot change this). The option is registrar specific ie if user1 changes the setting, user2 can change it back given that both users belong to the same registrar and have at least billing roles. 

- Admin provides e-invoice operator settings in admin area
- Registrar cannot select "e-invoice" option unless operator settings are present
- Registrar can subscribe/unsubscribe to/from billing subscription
- Every user of `billing` role can edit billing subscription at profile page in registrar area

Renewed Omniva spec: https://drive.google.com/open?id=19OqNolFk_REEAiS8C6498Ax4iA-CzZTM
Estonian e-invoice spec: https://media.voog.com/0000/0042/1620/files/Estonian_e-invoice_description_ver1.2_eng.pdf
https://www.pangaliit.ee/arveldused/e-arve

---

- [x] Registrar area UI
- [x] E-invoice generation gem (https://github.com/internetee/estonian_e_invoice)
- [x] Rake task
- [x] Rake task to convert reference numbers of existing registrars (https://github.com/internetee/registry/pull/998)
- [x] Refactor invoices
- [x] E-invoice gem integration
- [x] Test against Omniva test env

artur-beljajev:
/registrar/invoices/61/forward

artur-beljajev:
@vohmar or @ratM1n Are there any issues with invoicing?

vohmar:
@artur-beljajev in production not that we know of.

artur-beljajev:
@vohmar Is it clear what billing provider will be used?

vohmar:
* We have contract with Omniva for e-invoicing

artur-beljajev:
[...]

PR #998, opened 2018-10-02
Use Estonian reference number format instead of ISO 11649

artur-beljajev:
This ticket regenerates reference number of all existing registrars.

All previously generated numbers in `invoices.reference_no` column that begin with "RF" (ISO 11649) remain unchanged.

`bundle exec rake data_migrations:regenerate_registrar_reference_numbers` task should be run after `rake db:migrate` for proper deployment (the former applies data migration, whereas the latter applies structure migration). The task is idempotent, which means it will not regenerate numbers in new format if you run it repeatedly.

Verify:
- Reference number generation when adding a new registrar
- Bank transaction - invoice matching
- Previously issued invoices are matched (after running the above mentioned rake task)

Part of #329.


artur-beljajev:
@vohmar I have just realized, that the data migration fails, if registrar is invalid (invalid in terms of current business logic). What's the status with testing? I might change that part to only verify the conversion, ignoring all validations.

artur-beljajev:
@ratM1n Please run on prod the following queries:
`SELECT COUNT(*) FROM registrars WHERE reference_no IS NULL`
`SELECT COUNT(*) FROM invoices WHERE reference_no IS NULL`

PR #1010, opened 2018-10-16
Add handling of OPTIONS request to /api namespace

maciej-szlosarczyk:
* It allows access from anywhere via wildcard origin
* It sets the timeout to an hour
* It allows all standard HTTP verbs + OPTIONS


maciej-szlosarczyk:
@ratM1n, this should fix yesterday's problem with `OPTIONS` requests.

PR #1013, opened 2018-10-18
Refactor and improve invoices

artur-beljajev:
### Changes highlights:
- `runner 'Invoice.cancel_overdue_invoices'` in `schedule.rb` is changed to 
`rake 'invoices:cancel_overdue'`.
- `invoices.payment_term` database column is removed and its value is hardcoded in UI.
- `invoices.paid_at` is removed as unused (@ratM1n reported zero rows with any value in prod as of 2019-03-11)
- `invoices.due_date` column's type is now `date`.
- `Invoice#invoice_items` renamed to `Invoice#items` and `Invoice` interface to get a list of items is unified.
- Default date format in UI.
- Default translations are used.
- Tests improved.
- Specs converted to tests and removed along with factories.
- Database structure improved.

### Deployment:
- To deploy, `rake data_migrations:populate_invoice_issue_date` should be run _after_ `rake db:migrate`. This task populates `invoices.issue_date` column with the date from `invoices.created_at`.

All invoices should be valid for a data migration to be able to run. @ratM1n Reported zero invalid invoices on prod on 2018-10-18, but it may change. To get a list of them run `bundle exec rails c` and insert the following: https://gist.github.com/artur-beljajev/c3303eb3b39dcedf4a721548d0f3e6b7

### Verify:
  - Account top-up (total calculation)
  - Invoice cancellation (incl. automatic)
  - Payment-invoice matching
  - Directo integration
  - Changing `days_to_keep_overdue_invoices_active` setting in Admin portal

---


app/controllers/registrar/invoices_controller.rb
```diff
@@ -61,8 +61,8 @@ def normalize_search_parameters
 
       ca_cache = params[:q][:due_date_lteq]
       begin
-        end_time = params[:q][:due_date_lteq].try(:to_date)
-        params[:q][:due_date_lteq] = end_time.try(:end_of_day)
+        end_time = params[:q][:due_date_lteq]
+        params[:q][:due_date_lteq] = end_time
```
maciej-szlosarczyk: I don't understand this without the additonal methods, how is it different than:

```ruby
a = 1
b = a
a = b  # can be skipped
```

The rescue on lines 67-69 also seems weird now, the only error you might see is `NoMethodError` when `q` hash is `Nil`.

db/structure.sql
```diff
@@ -1052,7 +1052,8 @@ CREATE TABLE public.invoices (
     cancelled_at timestamp without time zone,
     total numeric(10,2) NOT NULL,
     in_directo boolean DEFAULT false,
-    buyer_vat_no character varying
+    buyer_vat_no character varying,
+    date date
```
maciej-szlosarczyk: I am not sure if this is the right choice. 

You need to know that it means the date the invoice was issued, and not due date, and it conflicts with SQL `DATE` keyword.

lib/tasks/data_migrations/populate_invoice_date.rake
```diff
@@ -0,0 +1,16 @@
...
+    Invoice.transaction do
+      Invoice.find_each do |invoice|
+        invoice_date = invoice.created_at.to_date
+        invoice.update!(date: invoice_date)
+
+        processed_invoice_count += 1
+      end
+    end
+
+    puts "Invoice processed: #{processed_invoice_count}"
```
maciej-szlosarczyk: Minor typo: `Invoices processed`, plural

app/controllers/registrar/invoices_controller.rb
artur-beljajev: I postponed refactoring that part of legacy code. There are some other things I don't understand as well.

db/structure.sql
maciej-szlosarczyk: >I agree that DATE itself is a reserved keyword, but in what way and what exactly conflicts?

It is a reserved keyword in all [3 recent SQL standards](https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html), which makes it generally not portable. In Oracle you cannot query such a column without escaping its name, it raises a parser exception.

> I didn't notice any obstacles using it. Even if you need to use plain SQL, you just need to quote it.

That is precisely one of the reasons why you should avoid doing it. Quoted values are case sensitive, making them prone to bugs caused by typos and key rollovers.

```sql
create table dates(DATE DATE NOT NULL, ISSUE_DATE DATE NOT NULL);

select * from dates;
+--------+--------------+
| date   | issue_date   |
|--------+--------------|
+--------+--------------+

SELECT "Date" from dates;
column "Date" does not exist
LINE 1: SELECT "Date" from dates
               ^
HINT:  Perhaps you meant to reference the column "dates.date".

SELECT ISSUE_Date from dates;
+--------------+
| issue_date   |
|--------------|
+--------------+
SELECT 0
```

> Invoice has date so date seems like a natural, self-explanatory name of a column.

It has 3 dates to be exact: the date when it was created, the date when it was sent to the customer and the date when it is due. Because of that, just date is not the best choice IMHO.

lib/tasks/data_migrations/populate_invoice_date.rake
artur-beljajev: Fixed.

db/structure.sql
artur-beljajev: OK, at least now I understand the point. However, I still see `date` as a better fit. I bet nobody will ask you to clarify, if you say just "Invoice date".

vohmar: @artur-beljajev please fix the reserved date keyword issue. You always fight against ambiguity so it is hard to understand why all of the sudden its OK even if it is error prone and raises specific problems as @maciej-szlosarczyk described.

artur-beljajev: I don't see a problem using `date`, which is a standard term for invoice.
https://www.google.ee/search?hl=en&as_q=&as_epq=invoice+date&as_oq=&as_eq=due&as_nlo=&as_nhi=&lr=&cr=&as_qdr=all&as_sitesearch=&as_occt=any&safe=images&as_filetype=&as_rights=
Omniva uses the same term: https://media.voog.com/0000/0042/1620/files/Estonian_e-invoice_description_ver1.2_eng.pdf
Our `estonian_e_invoice` gem also sticks to it.

I find `invoice.date` clearer. 99.9% of time we will use application layer (i.e. call `invoice.date`), instead of manually building queries (in which case you will need to quote it). I don't find `invoice.date` ambiguous.

Please elaborate, what problem it introduces in our case?

artur-beljajev: P.S. Another solution is to employ `alias_attribute :date, :issue_date`

vohmar:
has merge conflicts with master

vohmar:
It appears that the change to start using reserved keyword date is still in effect here. So this change is postponed until the conflict is resolved. Wil come back to this after auction center release.

artur-beljajev:
@vohmar Is there anything preventing this PR from being merged with master? Invoice date issue has been fixed.

vohmar:
@maciej-szlosarczyk please review if the issues mentioned have been resolved (i still see columns called date in db migrations).

maciej-szlosarczyk:
@vohmar: Locally it seems that all is correct. Maybe you're looking at type:

```ruby
add_column :invoices, :issue_date, :date
```

The column name is `issue_date`, `date` is postgres column type.

artur-beljajev:
I have one more change for this PR.

app/services/overdue_invoice_canceller.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(invoices: Invoice.overdue, delay: self.class.delay)
+    @invoices = invoices
+    @delay = delay
+  end
+
+  def self.default_delay
+    30.days
+  end
+
+  def self.delay
```
maciej-szlosarczyk: I guess by renaming this to `delay_in_days` we might get rid of operations on duration and treat them as integers, as does `Date`.

```ruby
def self.default_delay_in_days
  30
end

def self.delay_in_days
  Setting.days_to_keep_overdue_invoices_active || default_delay
end

File: app/models/registrar.rb [hunk 5]
```diff
@@ -51,8 +51,8 @@ def ordered
 
   def issue_prepayment_invoice(amount, description = nil)
     invoices.create(
-      due_date: (Time.zone.now.to_date + Setting.days_to_keep_invoices_active.days).end_of_day,
-      payment_term: 'prepayment',
+      issue_date: Time.zone.today,
+      due_date: (Time.zone.now + Setting.days_to_keep_invoices_active.days).to_date,
```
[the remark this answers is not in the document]
artur-beljajev: It was like this before. I will refactor it in a separate branch.

app/services/overdue_invoice_canceller.rb
artur-beljajev: The point was to just red rid of `_days` suffix and to introduce an abstraction (a delay). It allows to keep it in any units, be it a second, minute or year.

vohmar:
cannot test automatic invoice cancellation on staging. The error:
```
registry@staging:~/registry/current$ /bin/bash -l -c 'export PATH="$HOME/.rbenv/bin:$PATH";eval "$(rbenv init -)"; cd /home/registry/registry/current && RAILS_ENV=staging bin/bundle exec rake billing:cancel_overdue_invoices'
rake aborted!
Don't know how to build task 'billing:cancel_overdue_invoices' (see --tasks)
/home/registry/registry/releases/699/vendor/bundle/ruby/2.4.0/gems/rake-12.3.1/exe/rake:27:in `<top (required)>'
bin/bundle:3:in `load'
bin/bundle:3:in `<main>'
(See full trace by running task with --trace)
```

artur-beljajev:
> /home/registry/registry/releases/699/vendor/bundle/ruby/2.4.0/

@ratM1n Why we test against Ruby 2.4.0, whereas our project requires 2.4.5? https://github.com/internetee/registry/blob/master/.ruby-version

PR #1018, opened 2018-10-23
enable airbrake for all branches

teadur:

PR #1021, opened 2018-10-24
Fix airbrake configuration in test

maciej-szlosarczyk:
Fixes test suite problems introduced by #1018.

@teadur, please update all the environment with the following:

1. remove custom `config/initializers/errbit.rb` files, they would be overwritten.
2. Update `application.yml` with configuration values for errbit, as shown in  `application-example.yml` in this PR.
3. Pull this branch to the environment.
4. Verify that everything is in order by running `bundle exec rake airbrake:test`

I retained your patch on OpenSSL.


maciej-szlosarczyk:
You can now configure `airbrake-env` to be something different than Rails environment, for example `acceptance`, `testrant` or something else. There are still two reserved keywords: `development` and `test`. You should not use them if you want exceptions to be raised to Errbit.

If you do not, it will still default to Rails environment.

PR #1033, opened 2018-11-01
Remove `SyslogLogger` gem

artur-beljajev:
It is present in Ruby Standard Library since version 2.0
https://github.com/seattlerb/sysloglogger

Affects production config, needs verification.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,8 +8,6 @@ end if Bundler::VERSION < '2'
 source 'https://rubygems.org'
 
 # core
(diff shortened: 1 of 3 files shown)


vohmar:
won't deploy on staging:

```
rake aborted!
       NameError: uninitialized constant Syslog
```

artur-beljajev:
What command? Is it from staging? Currently `Syslog` is only included in production (excl. `staging`)

Issue #992, opened 2018-09-21
Registrant API: enable disclosing private user's name and email address


vohmar:
Option for private user to disclose their name and/or email address in EIS's web whois just like before GDPR.

If a person decides to disclose their private data the output for general public whois queries should be similar to business contact data

Name - disclosed publicly ie also seen on public whois queries
Email - Not Disclosed - Visit www.internet.ee for webbased WHOIS

Disclosing query must also enable undisclosing already user disclosed data - not possible to undisclose data available by default (ie business contacts' name and email)

Current sketch of registrant portal on how this functionality will be used by domain contacts:
https://projects.invisionapp.com/share/N4K6NII3JYG#/screens/285863083_Kontakti_Muutmine

artur-beljajev:
So only API and no UI? (which is being implement by a third-party, I guess?)

vohmar:
The Ui functionality will be included in the new portal all ready in development - no reason to spend time on current one. So yer the functionality in regard of this ticket here is only focusing on Registrant Rest-API.

vohmar:
yes, only Registrant Rest-API development in this ticket.

teadur:
Are we going to do #325  at the same time ?

artur-beljajev:
I guess it also involves changes in `whois` and `rest-whois` projects?

artur-beljajev:
@vohmar What should be returned in `email` field of registrant, tech/admin contacts in JSON version (http://rest-whois.test/v1/shop.test.json)?

artur-beljajev:
@vohmar https://github.com/internetee/rest-whois/pull/129#issuecomment-435413895

artur-beljajev:
> Disclosing query must also enable undisclosing already user disclosed data

Before I was sure we don't have such feature, but it seems we do https://github.com/internetee/registry/pull/1038 (needs confirmation).

vohmar:
run into an issue on staging with one of the contact objects. Getting noMethodError ehen I try to open the details of the object in admin 

https://st-adm.infra.tld.ee/admin/contacts/238421

```
NoMethodError in Admin::Contacts#show
Showing /home/registry/registry/releases/678/app/views/admin/contacts/partials/_general.haml where line #41 raised:

undefined method `admin_registrant_user_path' for #<#<Class:0x00007fc80050bcf0>:0x00007fc7e99216f8>
```

Extracted source (around line #51):
```
    # can be api user or some other user
    link_to(model.updator, ['admin', model.updator])
[...]

PR #1038, opened 2018-11-08
Remove unused JSON key

artur-beljajev:
Deploy along with https://github.com/internetee/rest-whois/pull/163 and https://github.com/internetee/whois/pull/40.

Previously `registry` app was generating `whois_record.json['disclosed']` key which wasn't used neither by `rest-whois` nor by `whois` project.

It seems it was meant for https://github.com/internetee/registry/issues/992 initially, but I see no benefit leaving it.

Existing records will still have this key (until regeneration over time), should not create any problem.

---
Currently all emails of admin/tech contacts are mixed in a single collection that looks as follows:

```
  "disclosed": [
    [
      "email",
      "john@inbox.test"
    ],
    [
      "email",
      "john@inbox.test"
    ],
    [
      "email",
      "john@inbox.test"
    ]
  ]
```

Is this what you wanted to reuse in https://github.com/internetee/registry/issues/992#issuecomment-438037334? If so, we will first need to parse current state of WHOIS database to distinguish registrant email from admin/tech contact emails, and then regenerate the whole WHOIS database with new structure. It's kinda error-prone.

@vohmar Needs your input.

### DIFF
--- a/app/models/whois_record.rb
+++ b/app/models/whois_record.rb
@@ -33,7 +33,6 @@ class WhoisRecord < ActiveRecord::Base
 
     registrant = domain.registrant
 
-    @disclosed = []
     h[:disclaimer] = disclaimer_text if disclaimer_text.present?
     h[:name]       = domain.name
     h[:status]     = domain.statuses.map { |x| status_map[x] || x }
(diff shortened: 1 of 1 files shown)


vohmar:
default behavior in whois with private registrants is hide data for registrant and all other contacts. Disclose functionality allows users to override that behavior by revealing their own name and email address - but only their own other contacts of the same domain registration are not affected.

So it is important to specify exactly what data of which contact are disclosed.

PS: user can decide whether to disclose name or email address or both.

artur-beljajev:
> default behavior in whois with private registrants is hide data for registrant and all other contacts
OK, this is how it's implemented at the moment.

> Disclose functionality allows users to override that behavior by revealing their own name and email address - but only their own other contacts of the same domain registration are not affected.
So it is important to specify exactly what data of which contact are disclosed.
PS: user can decide whether to disclose name or email address or both.

This just rephrases the original description of https://github.com/internetee/registry/issues/992 and doesn't answer the question above. So, I guess we can just discard what has been done before and still push this PR to prod.

PR #1046, opened 2018-11-30
Change reference number algo

artur-beljajev:
Fixed: 
- Bypass registrar validation


teadur:
@artur-beljajev from where the request to bypass validation originates / whats the reasoning ?

PR #1054, opened 2018-12-11
Regenerate whois schema.rb

maciej-szlosarczyk:


maciej-szlosarczyk:
Makes it easier to set up development environments.

No testing should be required. I should've done it long ago, mostly for informational purposes.

PR #1058, opened 2018-12-17
Support JSON only

artur-beljajev:
This effectively fixes `registry-992`, which is already in master.

Contact update endpoint in Registrant API needs to be completely retested.

### DIFF
--- a/app/controllers/api/v1/registrant/contacts_controller.rb
+++ b/app/controllers/api/v1/registrant/contacts_controller.rb
@@ -53,7 +53,8 @@ module Api
 
           if disclosed_attributes
             if contact.org?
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/registrant/contacts_controller.rb
```diff
@@ -53,7 +53,8 @@ def update
 
           if disclosed_attributes
             if contact.org?
-              error_msg = "Legal person's data cannot be concealed. Please remove this parameter."
+              error_msg = "Legal person's data is visible by default and cannot be concealed." \
```
artur-beljajev: @vohmar FYI

Issue #874, opened 2018-06-12
Auction integration


vohmar:
- [x] API for `auction_center`
- [x] EPP info/check basic tests
- [x] EPP support
- [x] Generate WHOIS records for domains `at auction`/`awaiting_payment`/`pending_registration`
- [x] `rest-whois` support
- [x] `whois` support

New settings:
- [x]  `release_domains_to_auction` (feature flag to switch from `deleteCandidate` procedure to auction mode)
- [x]  `auction_api_allowed_ips`

New rake task:
- [x] `domains:release` (needs to be run periodically)

Removed rake task:
- [x] `domain:discard`

https://docs.google.com/document/d/1cBYPIDuOnQzHJ40kJsvSZXS-wVphR0-nVJu-OQbfIxI/edit?ts=5bfc525c

https://github.com/internetee/registry/blob/42e8f86daeec44e2f7afbd48100bc376671c5e7d/config/application-example.yml#L150

Take into account existing discarded domains when going live (+ pending background jobs)

vohmar:
Auctionig process together with status updates on registry side works. Whois is still a problem.

PR #1062, opened 2019-01-02
Add basic tests before auction integration

artur-beljajev:
Does not affect production code.

Needed for https://github.com/internetee/registry/issues/874.


vohmar:
autotests fail with this branch

PR #1069, opened 2019-01-16
Auction integration

### LINKED ISSUE #874: Auction integration

artur-beljajev:
Closes #874


app/controllers/api/v1/auctions_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+        render json: Auction.started.map { |auction| serializable_hash(auction) }
+      end
+
+      def show
+        auction = Auction.find_by(uuid: params[:uuid])
+        render json: serializable_hash(auction)
+      end
+
+      def update
+        auction = Auction.find_by(uuid: params[:uuid])
```
maciej-szlosarczyk: Auction not found is not handled gracefully, it raises a `NoMethodError` in the  case statement below.

app/models/auction.rb
```diff
@@ -0,0 +1,63 @@
+class Auction < ActiveRecord::Base
+  enum status: {
+    started: 'started',
+    awaiting_payment: 'awaiting_payment',
+    no_bids: 'no_bids',
+    payment_received: 'payment_received',
+    payment_not_received: 'payment_not_received',
+    domain_registered: 'domain_registered',
```
maciej-szlosarczyk: We need additional state, `domain_not_registered`, which would functionally do the same thing as `payment_not_received`, create a new auction for the same domain name. @vohmar 

There's instrumentation already for that on auction center's side.

app/controllers/api/v1/auctions_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+        when Auction.statuses[:awaiting_payment]
+          auction.awaiting_payment!
+        when Auction.statuses[:no_bids]
+          auction.mark_as_no_bids
+        when Auction.statuses[:payment_received]
+          auction.mark_as_payment_received
+        when Auction.statuses[:payment_not_received]
+          auction.mark_as_payment_not_received
+        else
+          raise "Invalid status #{params[:status]}"
```
maciej-szlosarczyk: This is a RuntimeError, which results in a response with 500 HTTP code. The response should indicate that it's a client error, not a server one.

app/controllers/api/v1/base_controller.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails5_api_controller_backport'
+
+module Api
+  module V1
+    class BaseController < ActionController::API
+      private
+
+      def authenticate
```
maciej-szlosarczyk: We might need to add another layer of security, such as API key sent in Authorization headers. It can be static, defined in environment, and not backed by an actual registry user.

maciej-szlosarczyk: Authentication based on IP _only_ is pretty weak, they only validate the host, and not the actual user. For more info:

https://security.stackexchange.com/questions/4533/in-what-scenarios-is-relying-on-source-ip-address-as-a-security-control-acceptab

https://security.stackexchange.com/questions/13255/how-secure-can-ip-based-login-be

artur-beljajev: I don't argue that it's weak. My point was to keep it as is for the launch, given it is not accessible from Internet.

Perhaps we will need some message broker instead of API, such as rabbitmq...

PR #1075, opened 2019-01-30
Support id card v2018

artur-beljajev:
Awaits "Äriregister" problem resolution.

Required for https://github.com/internetee/registry/blob/594a8bbeaf5ca019072d382900719cd65515b9a0/test/integration/registrant_area/sign_in/id_card_test.rb#L5. It was the fastest way to proceed until component is refactored.


artur-beljajev:
Requires #1077

PR #1103, opened 2019-03-12
Remove duplicate translations

artur-beljajev:
Nothing should be affected.

@maciej-szlosarczyk Would be good to review in case I forgot something.


maciej-szlosarczyk:
I have no idea what I'm supposed to review here.

PR #1109, opened 2019-03-19
Revamp WHOIS for auctions

### LINKED ISSUE #1105: WHOIS does not return statuses for domains in auction process
1. It appears that after running domains:release rake task only some domains are written into whois_records table in registry db. No errors are returned anywhere.
2. auctions ending with no_bids status are not removed from whois_records table and remain there with AtAuction status
3. domain names that had successful bids and have status awaiting payment to not get update in whois_records and remain in whois with AtAuction status while they should have status "Pending regsitration"

artur-beljajev:
Closes #1105, #1111


vohmar:
merge conflict

PR #1122, opened 2019-03-22
Do not load the whole environment for rake tasks

artur-beljajev:
Verify `domains:release` task.


vohmar:
domains were released to auction but so were reserved and blocked domains that should not go to auction. Possibly this branch is missing the reserved/blocked fix.

artur-beljajev:
yes, since master went forward already. If the task itself runs, then it works.

Issue #253, opened 2016-11-18
ForceDelete: force_delete_at saved to db in server tz instead of utc


vohmar:
on setting forcedelete to a domain the force_delete_at value saved must be the utc value of the beginning of a day according to server timezone - ie 

"2016-12-18 22:00:00"

to represent 2016-12-19 00:00:00+02

but instead "2016-12-19 00:00:00" is saved.

artur-beljajev:
@vohmar As I see, it is still valid. Can we move this up?

artur-beljajev:
@vohmar It seems the same applies to `delete_at`. Can you check this?

vohmar:
delete_at gets its value from valid_to value +45 days. as valid_to value is in datetime format and timed to the end of the day according Estonian timezone (ie 2019-02-04 21:00:00) then delete_at has the same format and time 45 days later. Removing time here would case the domain to be release a day too early unless all delete_at values are incremented by 1.

this tickets pullrequest is merged and in production so I will close this ticket now.

PR #1124, opened 2019-03-22
Refactor force delete

### LINKED ISSUE #812: Force Deleted domains do not get deleteCandidate status on the last day
Domains in forceDelete process do not get deleteCandidate status assigned to mark the day of deletion and prohibit any changes to the registration. 

Expected behavior is identical in all cases - regular expiration process, domain delete epp request and force delete procedure - after the last 30 day period (redemption grace period) has passed (in case of normal expiration process since setting the onHold status after expiration warning period has passed, or since the domain delete request or setting Forcedelete) deleteCadidate status is set and que job is created to control the random delete time.

### LINKED ISSUE #1119: ForceDelete: domains are not set for release nor sent to the auction
domain:release rake task does not handle domains with forceDelete status set.

ForceDelete sets separate delete date for the domain and does not touch delete_at value. So if both are set the one that arrives earlier takes an effect.

Delete process ends the same way for all forms of releasing the domain - expiration, epp domain:delete or adminstrative forceDelete. If auction is disabled deletecandidate status is set and delayed job is created to release the domain name at random time within next approximately 24 hours. If auction is enabled then all the released domains including the ones with forceDelete set must be auctioned off

https://github.com/internetee/registry/blob/e4305e78656ad6610f5c85d9263eef93360a133d/app/models/domain_cron.rb#L82

artur-beljajev:
- Change `domains.force_delete_at` database column type to date,
rename to `force_delete_date`
- Save `force_delete_date` in application timezone
- `DomainCron.destroy_delete_candidates` runner is removed from
`config/schedule.rb` with `domains:release` rake task as a replacement
- Improve tests

Verify:
- `domains:release` rake task
- `force_delete_date` in WHOIS
- Updating domain statuses from Admin area, when a domain has `DomainStatus::SERVER_MANUAL_INZONE` status

Fixes #812, #253
Closes #1119

### DIFF
--- a/app/models/concerns/domain/force_delete.rb
+++ b/app/models/concerns/domain/force_delete.rb
@@ -12,8 +12,7 @@ module Concerns::Domain::ForceDelete
 
     preserve_current_statuses_for_force_delete
     add_force_delete_statuses
-    self.force_delete_at = (Time.zone.now + (Setting.redemption_grace_period.days + 1.day)).utc
-                             .beginning_of_day
+    self.force_delete_date = Time.zone.today + Setting.redemption_grace_period.days + 1.day
     stop_all_pending_actions
     allow_deletion
     save(validate: false)
@@ -22,7 +21,7 @@ module Concerns::Domain::ForceDelete
   def cancel_force_delete
     restore_statuses_before_force_delete
     remove_force_delete_statuses
-    self.force_delete_at = nil
+    self.force_delete_date = nil
     save(validate: false)
   end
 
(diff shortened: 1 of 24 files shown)


vohmar:
* if force delete is applied and some of the statuses that comes with forcedelete is already applied the domain ends up with two statuses of the same kind - there cannot be more than one
* if force delete is removed and some of the statuses included in the force delete were applied in before forcedelete, all of the statuses are removed. Removing force delete should restore the previous state of the domain. so if transfer prohibited was there before, it shoukd be there after removinf force delete as well.

artur-beljajev:
Is it something new so that only branch in question is affected or master also has those issues?

vohmar:
issue moved to separate tickets
https://github.com/internetee/registry/issues/1135
https://github.com/internetee/registry/issues/1136

PR #1140, opened 2019-04-02
Handle IDN domain names in EPP check and EPP create

maciej-szlosarczyk:
Discovered in auction integration, but also affects existing reserved and blocked domains.

### DIFF
--- a/test/integration/epp/domain/create/auction_idn_test.rb
+++ b/test/integration/epp/domain/create/auction_idn_test.rb
@@ -1,0 +1,282 @@
+# encoding: UTF-8
+require 'test_helper'
+
+class EppDomainCreateAuctionIdnTest < ApplicationIntegrationTest
+  def setup
+    super
+
+    @idn_auction = auctions(:idn)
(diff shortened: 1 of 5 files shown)


test/integration/epp/domain/create/auction_idn_test.rb
```diff
@@ -0,0 +1,282 @@
...
+    Domain.release_to_auction = true
+  end
+
+  def teardown
+    super
+
+    Domain.release_to_auction = false
+  end
+
+  def test_registers_domain_with_ascii_idn_cannot_be_registered_without_registration_code
```
artur-beljajev: I don't understand this. Forgot to delete `registers`?

vohmar:
blocked domain still affected, fix works for reserved and auctioned domains

test/integration/epp/domain/create/auction_idn_test.rb
maciej-szlosarczyk: Copy and paste error

PR #1143, opened 2019-04-03
Create whois records on domain release

maciej-szlosarczyk:
Fixes an issue from staging.

### DIFF
--- a/app/controllers/api/v1/auctions_controller.rb
+++ b/app/controllers/api/v1/auctions_controller.rb
@@ -52,7 +52,7 @@ module Api
       end
 
       def update_whois_from_auction(auction)
-        whois_record = Whois::Record.find_or_create_by(name: auction.domain) do |record|
+        whois_record = Whois::Record.find_or_create_by!(name: auction.domain) do |record|
           record.json = {}
         end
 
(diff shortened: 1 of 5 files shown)


vohmar:
on staging: three domains were released martin53.ee, martin54.ee and martin55.ee. After running domain:release martin53.ee returned not found in whois, the other two had at auction records in whois. All three had correct whois records before running domain:release. So the other two got new records created but 53 did not. Did number of redeploys and the result was consistent 53 never got new atauction record to whois, the other two always did.

vohmar:
the whois record for martin53.ee was created after ResultStatusUpdateJob after successful bid and end of the auction. So there is still something off with the domain:release task not creating the missing whois record.

maciej-szlosarczyk:
I have some idea what might go wrong there. Please, when you have time, try to test again. There should be an error raised when a whois record is not created.

maciej-szlosarczyk:
As it stands, this improves some previously butchered test names from #1140 and makes sure that the domain:release rake task does not fail if whois record does not exist (we saw that on staging).
Unfortunately the original issue remains unresolved.

PR #1146, opened 2019-04-04
Refactor contact archivation

### LINKED ISSUE #956: Destory orphans cron task does not archive unlinked contacts
Business requirement: all contacts that have not been linked to any domain for more than 6 months must be archived (removed from the list of active contact objects)

There is contact status linked that means that contact is associated with at least one domain. Linked status is assigned when contact is associated with a domain and the status is removed once the last domain association is removed.

For this purpose there is a cron job: Contact.destroy_orphans that should look for contacts with no relations to any registered domain name then determine the period contact has been unlinked and if that period is 6 months or more that contact object should be removed from contacts table (will remain in log_contacts table)

artur-beljajev:
- New task: `contacts:archive`
- Archive inactive contacts every midnight
- Change `log_domains.children` column's type to `jsonb`

@ratM1n 
- Notice changes in `schedule.rb` (https://github.com/internetee/registry/pull/1146/files#diff-b6de01cbd4b49ee8c835f6d988693033)
- A migration https://github.com/internetee/registry/pull/1146/files#diff-b3284f0eadfb18d72034542001018570 can take some time on production

Fixes #956


ratM1n:
`          <span class="error">Web application could not be started</span>
          <h3>Error ID:</h3>
          <span class="error">3105afb3</span>
          <h3>Details:</h3>
          <pre>undefined method 'months&apos; for nil:<wbr>NilClass (NoMethodError)
`

vohmar:
the process went on for 1,5 days (started about 16:30 at 23.07 and canceled at 10:30 at 25.07. Staging db has about 160000 contact objects and during that time none were actually archived.

This process in its current state is unusable as we would like to run it daily  - it cannot go on for more than couple of hours the most.

artur-beljajev:
It then needs deeper benchmarking, since I am unable to reproduce such long operation. That is definitely not OK.

app/models/inactive_contacts.rb
```diff
@@ -7,15 +7,9 @@ def initialize(contacts = Contact.archivable)
 
   def archive(verified: false)
     contacts.each do |contact|
-      log("Archiving contact: id(#{contact.id}), code(#{contact.code})")
-
+      puts "Archiving contact: id(#{contact.id}), code(#{contact.code})"
```
yulgolem, line 10: Is there any reason we are not using logger?

yulgolem (approved):

app/models/inactive_contacts.rb
karlerikounapuu: Logging is currently work in progress. Guess I asked for review too early 🙂

karlerikounapuu:
Use `bundle exec rake contacts:archive` to archive contacts. As the task is ultimately time consuming,  it's possible to start it again from where it left off via contact id. This can be done via `bundle exec rake contacts:archive[123]`, where `123` is the last `contact.id `which was processed (this can be obtained from tasks's `stdout`)

PR #1147, opened 2019-04-05
Fix emails' content

artur-beljajev:
Fixes content of "force delete" procedure emails. Preview of both HTML and text versions is available at `/rails/mailers/`.

Needs verification.


vohmar:
verified

PR #1149, opened 2019-04-05
Improve registrant change mailer

artur-beljajev:
- DRY templates
- Add preview

Verify emails' content (signature & confirmation link parts) and sending of confirmation email.


vohmar:
emails are sent and content is almost the same as it was, small changes in spacing.
Incorrect url is still there in the staging emails, but the problem was there before

artur-beljajev:
URL problem you mentioned is addresses by https://github.com/internetee/registry/pull/1151.

PR #1151, opened 2019-04-07
Revamp `action_mailer` config

artur-beljajev:
`registrant_url` setting in `application.yml` was misnamed. In fact, it had nothing to do with Registrant Portal. The intention is to provide default URL options for `action_mailer`, therefore `registrant_url` is removed, and new settings are introduced instead:

- `action_mailer_default_protocol`
- `action_mailer_default_host`
- `action_mailer_default_port`

@vohmar 
Verify links in emails. The settings above affect what protocol, host and port is used.

@ratM1n 
This change should fix the problem you had on staging. Previously `port` part of URL was ignored. Now there is a separate setting for this.

Note that `action_mailer_default_host` is required, whereas the rest is optional.


vohmar:
the confirmation email still contains wrong confirmation link url eve though application.yml has correct settings. But apparently this pr does not attempt to fix that bug.

artur-beljajev:
@vohmar Actually it does.

PR #1162, opened 2019-04-11
Do not send notification email if a contact does not act as a registrant on EPP contact:update

artur-beljajev:
Based on https://github.com/internetee/registry/pull/1160.

Fixes:
- Bug, when an email is sent if contact update fails under some circumstances
- #1161

Verify EPP `contact:update`


vohmar:
merge conflict

PR #1159, opened 2019-04-10
Refactor invoice PDF generation, download and delivery

artur-beljajev:
- Remove `Que::Mailer` (#895)
- Extract controllers
- Extract translations
- Convert HAML to ERB
- Add mailer preview
- Improve UI
- Remove unused routes
- Add tests

PR #1168, opened 2019-04-14
Encode domain parts of all addresses in mailers as punycode

artur-beljajev:
- Fixes bug when non-ascii emails are not allowed in `InvoiceMailer` (introduced in https://github.com/internetee/registry/pull/1159/files).
- Adds non-ascii email's domain part support for `from`, `cc` and `bcc` fields.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,10 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
   append_view_path Rails.root.join('app', 'views', 'mailers')
   layout 'mailer'
-
-  def format(email)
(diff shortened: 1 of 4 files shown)


vohmar:
merge conflict

vohmar:
the punycode domainnames need to be converted back to UTF whenever the domainname is displayed to the user. Ie registrant email change notification email contains the email address in the email body, registrant portal contact view, admin portal, registrant portal etc

vohmar:
apparently the problem brought out is not related to this PR and is already present in master. I will create separate ticket for this issue about IDN domain name support in user interfaces.

PR #1170, opened 2019-04-17
Add custom `From` header for "force delete" email

### LINKED ISSUE #1164: ForceDelete: option to add reply-to address
We need to get feedback from forcedelete domain registrants so we need an option to set reply-to address for forcedelete notification emails.

artur-beljajev:
New config: `action_mailer_force_delete_from`

Closes #1164


vohmar:
merge conflict

PR #1173, opened 2019-04-19
Change domains delete at to date

artur-beljajev:


vohmar:
the rake task ads a day to every value in the delete_date columnt - that is incorrect.

Issue #1031, opened 2018-10-31
VAT should always apply


artur-beljajev:
> Closer analysis revealed that there is no case in our business where vat is not applied. It is always applied the question is only whether it is 0% or more (no vat and 0% vat are different things legally).
> 
> So the initial specification stating VAT is not applied nor displayed on the invoice is incorrect that vat should be always applied and in the case described the value is 0%.

https://github.com/internetee/registry/issues/1006#issuecomment-434629656

Needed for #329

artur-beljajev:
```
registry=# select count(*) from registrars where vat_rate is null
registry-# ;
 count 
-------                                                                                                                                                            
    55                                                                                                                                                             
(1 row)                                                                                                                                                            
                                                                                                                                                                   
registry=# select count(*) from invoices where vat_rate is null;
 count 
-------
    41
(1 row)
```

artur-beljajev:
@vohmar See https://github.com/internetee/registry/pull/1174 and https://github.com/internetee/registry/pull/1175.

artur-beljajev:
https://github.com/internetee/registry/pull/1175

vohmar:
Adding credit fails. To reproduce:

* foreign registrar with vat nr filled (vat rate = null)
* add credit ends with validation error - Validation failed: VAT rate is not a number

![image](https://user-images.githubusercontent.com/7734747/57933932-28cf2580-78c7-11e9-8884-bbc30fac14f1.png)

artur-beljajev:
It is not ready yey... You probably forgot to reassign it back to me since the last issue...

PR #1175, opened 2019-04-21
Populate invoice VAT rate

artur-beljajev:
Deploy before and separately from #1031.

To deploy run `data_migrations:populate_invoice_vat_rate` task.


vohmar:
Running data_migrations:populate_invoice_vat_rate rake task resulted witn report "Invoices processed: 1", No invoice number nor id was specified and the single invoice with null as vat_rate value had still null as the vat rate in db.

the rake task output should also return some indicator  which invoices were changed in addition to actually populating vat_reate values

artur-beljajev:
Provide me addition info about invoices with `vat_rate = NULL` that were not processed.

vohmar:
staging id = 68

PR #1186, opened 2019-05-06
Remove mail templates feature as buggy and unused

artur-beljajev:
Nothing to test.


artur-beljajev:
@vohmar Is there anything that prevents this change from being merged?

PR #1191, opened 2019-05-09
Remove invoices seller and buyer address parts

artur-beljajev:


artur-beljajev:
Invoice address parts cannot be dropped, since at least e-invoice feature requires them.

PR #1202, opened 2019-05-16
Improve registrar area account view

artur-beljajev:
- Let registrar users change registrar account

Needed for https://github.com/internetee/registry/issues/329.


vohmar:
the registrar data validation seems too much in profile data updates because registrar's user cannot fix the data so for example validation fails like Reg no is missing, Contact e-mail is missing, Address street is missing, Address zip is missing, Address city is missing, Address state is missing, Reference number is invalid seem unjustified for billing address update.

artur-beljajev:
This is normal validation, so it is expected.

vohmar:
blocked by https://github.com/internetee/registry/issues/1206

PR #1203, opened 2019-05-16
Add registrar IBAN

artur-beljajev:
Requires https://github.com/internetee/registry/pull/1202, https://github.com/internetee/registry/pull/1204.
Needed for https://github.com/internetee/registry/issues/329.

### DIFF
--- a/app/controllers/admin/registrars_controller.rb
+++ b/app/controllers/admin/registrars_controller.rb
@@ -73,6 +73,7 @@ module Admin
                                         :vat_rate,
                                         :accounting_customer_code,
                                         :billing_email,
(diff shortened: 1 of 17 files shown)


vohmar:
* please replace the comment in profile view "Required for e-invoice" with "Required for sending e-invoices to the bank" - it is not needed if registrar wants e-invoice just to their ERP or email system.
* Saving IBAN number in registrar portal returns error:
image
@staging and probably caused by some invalid data in registrar profile

artur-beljajev:
No. 1 is fixed.
No. 2 should be fixed manually, since `reference_no` is auto-generated required value.

PR #1212, opened 2019-05-30
Correct offences reported by Rubocop

artur-beljajev:
Using `rubocop --safe-auto-correct`.

I briefly checked all changes manually as well, seems OK. There is a lot of tedious, repetitive and manual work otherwise, so I think this change is worth it.


artur-beljajev:
It turns out that even safe cops are not safe, so it's too risky to apply this change.

PR #1243, opened 2019-06-26
Issue e invoice along with invoice

### LINKED ISSUE #1222: E-invoice support
related to https://github.com/internetee/registry/issues/329

while in the ticket #329 auto invoicing the focus is on automatic invoice generation and delivery on set parameters.

But starting from the 1st of July all Estonian government institution accept only e-invoices. While auto invoicing concentrates on delivering the invoice to the bank so registrar can configure automatic payments in gov case we need to send the e-invoices to the e-invoice center. For this user has to have an option to request an invoice to be sent to e-invoice center. The option should enable user to set the e-invoice as a default form and delivery method for invoice so whenever new invoice is created it is instantly sent as e-invoice to user defined destination (bank/e-invoice center)

artur-beljajev:
Closes #1222.

Test invoice delivery channel.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -17,7 +17,7 @@ GIT
 
 GIT
   remote: https://github.com/internetee/e_invoice.git
-  revision: 60084e2035dae0a8bce6eda960f8f5a64c4da604
+  revision: 917318bd546322408b83567745375c998619c926
   branch: master
   specs:
     e_invoice (0.1.0)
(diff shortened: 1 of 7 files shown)


vohmar:
creating an invoice fails
image

vohmar:
resolved the token issue
but new and better one:
image

vohmar:
The internetbank channel is present in XML even if the iban for the registrar is missing

Issue #1078, opened 2019-02-02
Upgrade bundler to 2.x


artur-beljajev:
https://bundler.io/v2.0/guides/bundler_2_upgrade.html

> Version autoswitching means that Bundler will automatically run the correct 1.x or 2.x version based on your lockfile

Existing app cannot be updated via `bundle update --bundler`, since Rails 4 supports Ruby >= 1.9.3, whereas bundler 2.x does not support Ruby < 2.3.
https://github.com/rails/rails/blob/v4.2.11/rails.gemspec#L10
https://github.com/rails/rails/pull/34925

Rails 5 does not have such limitation.
https://github.com/rails/rails/blob/master/rails.gemspec#L36

---
Run on every deployment:

1. `gem update --system` 
Updates RubyGems
2. `gem install bundler` 
Updates bundler
---
Blocked by #377

teadur:
Tests seem to run fine https://travis-ci.org/internetee/registry/builds/580155989

artur-beljajev:
Not relevant anymore.

artur-beljajev:
https://github.com/internetee/registry/pull/1275#issuecomment-527408401

teadur:
Fixed by #1280

PR #1275, opened 2019-09-03
bundle with 2x

teadur:
Closes https://github.com/internetee/registry/issues/1078.


artur-beljajev:
https://github.com/internetee/registry/issues/1078

artur-beljajev:
Upgrade process is explained here in detail: https://github.com/internetee/registry/issues/1078#issue-405999019

teadur:
@artur-beljajev
The build seems to work fine to me, i don't know why ruby 2.3 would be problem when we are already on 2.4x

artur-beljajev:
@teadur Indeed, it seems I missed Ruby upgrade somehow 😕. So everything is OK.

One more thing: it's worth adding this line https://github.com/internetee/auction_center/blob/master/.travis.yml#L15 as well (as stated in Bundler upgrade doc).

PR #1280, opened 2019-09-03
Upgrade bundler and extend travis

### LINKED ISSUE #1078: Upgrade bundler to 2.x

teadur:
Use bundler version 2.x and  improve travis configuration
Closes #1078 .

Seems to work in staging but should be deployed with caution ( registrant portal  first for example )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: ruby
+rvm:
+  - 2.4
+  - 2.5
+  - 2.6
 cache: bundler
 env:
   - DB=postgresql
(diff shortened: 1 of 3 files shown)


artur-beljajev:
Can you squash related commits into single one?

teadur:
Yep, will be done on merge

PR #1284, opened 2019-09-03
Remove `mod_epp` docs

artur-beljajev:


vohmar:
shouldn't there be now new documentation regarding epp proxy - how to deploy and configure?

artur-beljajev:
I don't know... Basically it is covered here https://github.com/internetee/epp_proxy. EPP interface might be totally optional at all, so maybe we could just put some link there...

artur-beljajev:
@vohmar Reminder.

PR #1285, opened 2019-09-03
Update outdated doc

artur-beljajev:


vohmar:
instead of removing the doc it should be updated with information on how the setup and running of tests is done now.

artur-beljajev:
It is run using default `bin/rake test` and `bin/rspec` commands. Are you sure it needs to be specified? Our project anyway assumes some familiarity with Rails basics, so I don't think we need to cover such topic in detail.

artur-beljajev:
@vohmar Can we solve this problem comprehensively in a separate ticket? It makes no sense to update this doc in its current version. Nothing is helpful there.

yulgolem:
Updated testing.md documentation file to reflect our current testing approach using docker images.

PR #1295, opened 2019-09-09
Follow ruby name convention

teadur:


teadur:
Cosmetic changes, nothing to test.
Please squash and merge once review is done.

teadur:
> Nice change, but it should be `original_file_name.ext.sample`.
> Example: https://github.com/internetee/auction_center/blob/master/config/database.yml.sample
 should be okey now :)

Issue #960, opened 2018-08-30
Update `setup/bin`


maciej-szlosarczyk:
Instead of using `bin/setup` to set up new environment, we rely on documentation in readme. We can remove that, and use `bin/setup` to simplify the onboarding process for new developers.

Will also fix #189

artur-beljajev:
And not only for new developers, but also for existing ones who sets up a new instance.

PR #1296, opened 2019-09-09
Rails bin/setup with rbenv install

teadur:
First part of fixing #960, install build-run dependencies with bin/setup.
Should we solve anything else in this script ?


teadur:
New codeclimate issues are result of rubocop-i version increase not my code changes, because of that i think it makes more sense to merge the PR and work on the new issues gradually not in this PR.

PR #1331, opened 2019-09-18
Fix EPP session timeout

### LINKED ISSUE #711: EPP session timeout does not end sessions
There are millions of sessions still up in registry db indicating that session timeout is not ending idle sessions.

Expected behavior is that all sessions that are idle/inactive are ended after predefined period of time and the session record is deleted from the sessions table.

artur-beljajev:
@ratM1n @teadur 
- There is a new rake task `epp:clear_expired_sessions`. It should be scheduled to run periodically, depending on how often we want to purge expired sessions. Needed to be run manually on deploy, or otherwise nobody will be able to login.
- New setting https://github.com/internetee/registry/pull/1331/files#diff-fc5026eef1f4752ef71315b36727e708R161

Fixes #711

### DIFF
--- a/app/controllers/epp/base_controller.rb
+++ b/app/controllers/epp/base_controller.rb
@@ -368,7 +368,7 @@ module Epp
     end
 
     def session_timeout_reached?
-      timeout = 5.minutes
+      timeout = ENV['epp_session_timeout_seconds'].to_i.seconds
       epp_session.updated_at < (Time.zone.now - timeout)
     end
 
--- a/app/controllers/epp/base_controller.rb
+++ b/app/controllers/epp/base_controller.rb
@@ -10,7 +10,8 @@ module Epp
     before_action :latin_only
     before_action :validate_against_schema
     before_action :validate_request
-    before_action :update_epp_session, if: -> { signed_in? }
+    before_action :enforce_epp_session_timeout, if: :signed_in?
+    before_action :iptables_counter_update, if: :signed_in?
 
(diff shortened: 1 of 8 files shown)


teadur:
@artur-beljajev why does not running the cleanup break logging in ?
Whats the behviour when i just restart the application ?

artur-beljajev:
@teadur

- Why any logging is needed here?
- Behavior is not dependent on restart

vohmar:
this branch returns this error on running passenger:
`[Mon Oct 07 10:38:20 2019] [crit] [client 192.168.2.45] configuration error:  couldn't perform authentication. AuthType not set!: /epp/session/hello`

this is from Apache log - starting application fails.

artur-beljajev:
> [Mon Oct 07 10:38:20 2019] [crit] [client 192.168.2.45] configuration error: couldn't perform authentication. AuthType not set!: /epp/session/hello

Something is misconfigured on server side. Show me EPP backend Apache settings.

yulgolem:
Fixed some style & tests, let's check this.

karlerikounapuu (approved):

vohmar:
* epp:clear_expired_sessions rake tasks clear epp_sessions table, while it should only remove expired sessions (where now - updated_at is more than epp_session_timeout_seconds)

yulgolem:
The timeout variable was not set in the deploy script.
Added default timeout value of 300 sec, added test to see if task destroys only expired sessions.

PR #1346, opened 2019-10-08
Fix invoice pdf url display

### LINKED ISSUE #1188: Registrar website in invoice PDF is visible only if phone is present
https://github.com/internetee/registry/blob/master/app/views/invoice/pdf.haml#L210

teadur:
Noting to test.
Closes #1188


teadur:
> It's worth doing any reformatting (removing spaces in this case) in a separate commit.
Okey, i didnot notice that  its done in separate commit before henche didnot consider it as problem.
But will separate changes like that in future if you think its better like that.

app/views/invoice/pdf.haml
```diff
@@ -207,7 +207,7 @@
               %dt= t(:phone)
               %dd= @invoice.buyer_phone
 
-            - if @invoice.buyer_phone.present?
+            - if @invoice.buyer_url.present?
```
artur-beljajev: I think it makes sense to get rid of this check completely. Neither `admin` or `registrar` area invoice views have it. https://github.com/internetee/registry/blob/acab58b71cf367754e378489fc1d8a42724e6941/app/views/registrar/invoices/partials/_buyer.haml#L20
@vohmar is OK with this.

artur-beljajev: I also had an idea to rename it to `buyer_website` to be more generic, but this could be done later.

teadur: You are talking about check but you are not referencing one, you want me to remove the if statements from the haml ? I think this will end with nil errors

artur-beljajev: check = `if` statement on line no. 210.

teadur: As you wish, removed the if's.

teadur:
@artur-beljajev should be better now

teadur:
@artur-beljajev you have had 3 days to look over it, i assume its okey will merge today.

app/views/invoice/pdf.haml
```diff
@@ -195,25 +195,20 @@
             %dt= t(:reg_no)
             %dd= @invoice.buyer_reg_no
 
-            - if @invoice.buyer_address.present?
-              %dt= t(:address)
-              %dd= @invoice.buyer_address
+            %dt= t(:address)
```
artur-beljajev: I would remove custom translation key sitting at the root namespace (`address`), and use built-in `Invoice.human_attribute_name.buyer_address`. Or even better `Invoice::Buyer.address.human_attribute_name :address`.

artur-beljajev:
@teadur For some reason I even didn't see a notification for this PR. Perhaps due to
> teadur dismissed artur-beljajev’s stale review 2 days ago
this?

app/views/invoice/pdf.haml
teadur: Went with Invoice::Buyer.address.human_attribute_name :address.

teadur:
Seems better to me then the original situation, maybe we could get it merged now ?

yulgolem (approved):

PR #1347, opened 2019-10-08
Remove unused columns from users table.

### LINKED ISSUE #264: Consider removing "users.csr" and "users.crt" db columns
Seems that app uses "certificates" table instead

teadur:
In production there is no data in these column, nothing to test.
Closes #264

### DIFF
--- a/db/migrate/20191008024334_remove_csr_crt_from_users.rb
+++ b/db/migrate/20191008024334_remove_csr_crt_from_users.rb
@@ -1,0 +1,6 @@
+class RemoveCsrCrtFromUsers < ActiveRecord::Migration
+  def change
+    remove_column :users, :csr, :text
+    remove_column :users, :crt, :text
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20191008024334_remove_unused_from_users.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveUnusedFromUsers < ActiveRecord::Migration
```
artur-beljajev: Might be good idea to name it `RemoveUnusedColumnsFromUsers`. Or to follow existing convention: `RemoveColumn1Column2FromTableName`.

teadur: Should be better now.

teadur:
I think it makes sense to squash and merge either way, so whoever pushes it to master can do it.

artur-beljajev:
@teadur

I have squashed them into single one. Also edited commit message. https://chris.beams.io/posts/git-commit/ might be a good read on best practices.

Make sure `structure.sql` does not have unwanted changes. Newline at the end was missing.

Issue #247, opened 2016-11-15
Consider removing outdated docs


artur-beljajev:
https://github.com/internetee/registry/blob/master/doc/models_complete.svg
https://github.com/internetee/registry/blob/master/doc/controllers_complete.svg

vohmar:
@artur-beljajev no, these docs need to be updated, not removed

artur-beljajev:
Why do we need them?

vohmar:
it explains how the system works - data models have been a big help in the past. Documentation is good in case things are not self-explanatory.

artur-beljajev:
ok, then we probably need to automate this process somehow, since above mentioned docs can get outdated extremely easily (as they are now).

vohmar:
yes, please do. These schemas were automatically generated by the initial developers

artur-beljajev:
I submit removal of № 1, 2. Concerning № 3, they give no benefit, IMHO. They don't explain _how_ the system works, they just provide raw data of database columns and model relationship, which are anyway available by just looking at DB or models directly. Instead of investing time to support docs, I would invest this time in making the code more understandable.

If you still prefer to automate this, I kindly ask to explain the benefits we gain if we do this.

vohmar:
no1 might be necessary for modular architecture
no2 seems necessary as it guides the setup of the autotest environment.  
no3 db complete model must be kept updated
no4 update as necessary

artur-beljajev:
```
All registry demo data can be found at:
db/seeds.rb
```
is also outdated

vohmar:
just as a reminder we need to update these documents in addition to other doc existing and mentioned in this ticket (added links to rejected PRs as reference):

* database diagrams (https://github.com/internetee/registry/pull/1073)
* mod-epp doc needs to be replaced with epp-proxy doc (https://github.com/internetee/registry/pull/1284)
* autotest setup (https://github.com/internetee/registry/pull/1285)

there are gems used to update documentation previously. Some of these gems are listed here:
* https://github.com/internetee/registry/pull/1358/files

yulgolem:
2, 4 updated by https://github.com/internetee/registry/pull/1284

yulgolem:
3 there is no need to store in project, RubyMine generates them on the fly (https://www.jetbrains.com/help/ruby/working-with-diagrams.html#class_diagram).

PR #1358, opened 2019-10-11
Remove unused gems

artur-beljajev:
We don't use them and they seem to be effectively abandoned.

Also, removing `autodoc` gem allows us to fully get rid of `rspec` dependency.


artur-beljajev:
@vohmar Is there anything that prevents this from being merged?

vohmar:
these gems are used for automatically updating / generating documentation

teadur:
Seems to be related to #247

yulgolem:
sdoc gem will be substituted for yard gem.

Issue #377, opened 2017-02-21
Upgrade to Rails 5.0


artur-beljajev:

teadur:
Currently upgrade to Rails5 is blocked by paper_trail #426

artur-beljajev:
It is also blocked by lack of decent test coverage. For now this ticket is just to collect all feature/fixes I might need.

artur-beljajev:
https://hashrocket.com/blog/posts/how-to-upgrade-to-rails-5

artur-beljajev:
@maciej-szlosarczyk FYI

artur-beljajev:
https://github.com/rails/rails/pull/13157

artur-beljajev:
https://github.com/internetee/registry/pull/868

artur-beljajev:
https://github.com/ruby-i18n/i18n/releases/tag/v1.1.0

artur-beljajev:
https://github.com/rspec/rspec-rails/issues/1530#issuecomment-177883299

artur-beljajev:
`paper_trail` support for Rails 5 has been added in version 5.0.0 (https://github.com/paper-trail-gem/paper_trail/blob/master/CHANGELOG.md#500-2016-05-02)

artur-beljajev:
`constraints: {:id => /[^\/]+(?=#{ ActionController::Renderers::RENDERERS.map{|e| "\\.#{e}\\z"}.join("|") })|[^\/]+/}`

vohmar:
* Registrar - Contact create and update fails. Result:
![image](https://user-images.githubusercontent.com/7734747/68020242-910f9900-fca6-11e9-8b2b-4365b28a217a.png)
Query does not reach registry - there is no record in epp log
* EPP  - contact update resulted in 2400 command failed (https://st-adm.infra.tld.ee/admin/epp_logs/76985). Update fails for contacts that are linked to a domain. Unlinked contacts update OK.

artur-beljajev:
> Registrar - Contact create and update fails

create, update and delete controller actions are fixed.

vohmar:
* contact create, update and delete are OK

Admin:
* Domain versions do not open (not all domains have this problem, but a lot on staging do):
![image](https://user-images.githubusercontent.com/7734747/68046579-e2d71400-fce4-11e9-8c38-f27953951e79.png)
* account activities return error;
[...]

PR #1360, opened 2019-10-12
Improve domain and contact list download

artur-beljajev:
Affects:
- Domain list CSV download and search in registrar area
- Contact list PDF/CSV download and search in registrar area
- Domain list PDF/CSV download and search in registrant area

Slightly changes UI.

Needed for https://github.com/internetee/registry/issues/377.


vohmar:
* A bit less vague comments are appreciated: ie "download dropdown replaced with buttons" instead of "slightly changed UI"
* Charactercode error - Estonian special characters are not supported in downloaded PDF files (csv is OK)
image

artur-beljajev:
Nice catch, fixed.

PR #1365, opened 2019-10-15
Remove `money` gem

artur-beljajev:
Triggered by https://github.com/internetee/registry/issues/377, but I thought it would anyway be a good idea to get rid of additional dependency (which we don't actually use to the full extent) with quite little effort.

Affects billing and Directo integration.

Converts column type https://github.com/internetee/registry/pull/1365/files#diff-6979bcc7d7cf39dc881a09e613b83c9fR3.

Removes deprecation message on Rails 5:

> You are using the default localization behaviour that will change in the next major release. Find out more - https://github.com/RubyMoney/money#deprecation

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,6 @@ gem 'grape'
 # registry specfic
 gem 'isikukood' # for EE-id validation
 gem 'simpleidn', '0.0.7' # For punycode
-gem 'money-rails'
 gem 'data_migrate'
 gem 'whenever', '0.9.4', require: false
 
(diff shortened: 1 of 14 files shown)


vohmar:
please check your conversions!!!

vohmar:
nothing has changed - 80 still converts to 0.00. Please test before assigning!

artur-beljajev:
I asked you to provide more details in Slack.

vohmar:
more details with examples were in the chat - cents are incorrectly converted to euros loosing all the fractions. This is low priority issue, now on hold

vohmar:
with money gem all monetary values in db are kept in cents. This branch removes money gem and converts monetary values from cents ie integers to floats (euros and cents).
example values in the db before deploying this banch 80, 140
same values after deploying this branch 0.00, 1.00

PR #1378, opened 2019-10-19
Do not handle EPP contact:renew

### LINKED ISSUE #1362: EPP contact:renew incorrect response
Currently `Unimplemented command` is returned. The specification states:

>  Renewal semantics do not apply to contact objects, so there is no
   mapping defined for the EPP <renew> command.

https://tools.ietf.org/html/rfc5733#section-3.2.3

Consider returning `Unknown command` instead.

artur-beljajev:
I think it makes no sense (or rather impossible) to handle such command on registry app level, since there is just no such command/request defined (it's not even listed in [EPP XML schema](https://github.com/internetee/registry/blob/7c1e030f453a0bf2092bb066b5173b917fd4493e/lib/schemas/contact-ee-1.1.xsd#L27) which we validate against, so it will fail anyway). So I just removed the code that handled it altogether. App should just fail and return internal error back to EPP proxy. I am not sure we need to handle this on proxy level either, since it is not expected that someone will try to send such request.

Related to #1362

### DIFF
--- a/app/controllers/epp/contacts_controller.rb
+++ b/app/controllers/epp/contacts_controller.rb
@@ -70,12 +70,6 @@ module Epp
       end
     end
 
-    def renew
-      authorize! :renew, Epp::Contact
-      epp_errors << { code: '2101', msg: t(:'errors.messages.unimplemented_command') }
-      handle_errors
-    end
-
     def transfer
       authorize! :transfer, Epp::Contact
       epp_errors << { code: '2101', msg: t(:'errors.messages.unimplemented_command') }
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -63,7 +63,6 @@ class Ability
     can(:create, Epp::Contact)
     can(:update, Epp::Contact) { |c, pw| c.registrar_id == @user.registrar_id || c.auth_info == pw }
     can(:delete, Epp::Contact) { |c, pw| c.registrar_id == @user.registrar_id || c.auth_info == pw }
-    can(:renew,  Epp::Contact)
     can(:transfer,  Epp::Contact)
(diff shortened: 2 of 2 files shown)


vohmar:
I do agree that handling this scenario specifically is unnecessary, but at the same time I do not agree that unhandled internal error is better here - 2000 unknown command seems better aproach for any correctly structured request with unknown command - Exceptions are in my mind meant for unexpected result not unsupported input.

vohmar:
this is current result in registrar portal. Needs some formatting fixes:
image

artur-beljajev:
> correctly structured request

That's the point I tried to explain above: it is _not_ a valid EPP request according to our XML schemas.

vohmar:
we provide means to validate requests against schemas for registrars but not all are doing this. Yes this is not normal to send in invalid requests but that does not mean it never happens either in testing purposes, out of not knowing better or negligence. And we never had this kind of broken html gibberish responses before - even if we return internal error we should have decency to do in properly formatted way. 

But again internal errors are unexpected results that we would like to catch and most probably start handling in some way to avoid situations where exception log is full of so called known errors that we ignore.

artur-beljajev:
> we provide means to validate requests against schemas for registrars but not all are doing this
XML schema acts as a documentation. At the same time it is a contract that must be obeyed.

> does not mean it never happens either in testing purposes
Testing of what? There is no renew command for a contact object. I still can't imagine someone who decides to send it.

If you need to handle it on registry app side, we need to alter our schemas and put `renew` here https://github.com/internetee/registry/blob/7c1e030f453a0bf2092bb066b5173b917fd4493e/lib/schemas/contact-ee-1.1.xsd#L27.

yulgolem:
ATM renew requests are checking against the XML schema and response is 2001:syntax_error. If we'll remove path/controller action, the response will be 404 instead of EPP response, so closing this PR.

PR #1380, opened 2019-10-19
Fix typo

### LINKED ISSUE #371: Fix typo in setting name
`Setting.request_confrimation_on_registrant_change_enabled` (notice "conf**ri**mation").

artur-beljajev:
Rename `request_confrimation_on_registrant_change_enabled` setting to
`verify_registrant_change`.

@internetee/sysadmins `data_migrations:rename_setting` rake task must be run manually in order to remove old setting and create a new one (it will work anyway since the default value of a new one is the same we use, but it's better to persist the changes, and this is what this task does).

Fixes #371

### DIFF
--- a/app/models/epp/domain.rb
+++ b/app/models/epp/domain.rb
@@ -480,7 +480,7 @@ class Epp::Domain < Domain
     same_registrant_as_current = (registrant.code == frame.css('registrant').text)
 
     if !same_registrant_as_current && errors.empty? && verify &&
-       Setting.request_confrimation_on_registrant_change_enabled &&
+       Setting.verify_registrant_change &&
        frame.css('registrant').present? &&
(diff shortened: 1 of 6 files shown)


vohmar:
the result does not seem to be better. Its inconsistent with other settings and also seems to need now more clarification on the purpose of the newly renamed setting

compare:
Request confirmation on registrant change vs verify registrant change - first one is quite self explanatory the second one is not.

artur-beljajev:
Do you mean UI? I find `request_confirmation_on_registrant_change` too verbose. The process is verification. What gets verified? Registrant change process.

artur-beljajev:
I would clarify it in UI, if you think it's needed.

vohmar:
fixed by #1681

PR #1405, opened 2019-11-02
Tune kaminari gem

artur-beljajev:
Verify pagination. Should fix pagination-related problems described in https://github.com/internetee/registry/issues/377.


vohmar:
domain versions open ok.

PR #1411, opened 2019-11-06
Do not show `dnskey` and `nameserver` errors in UI

artur-beljajev:
Fixes issues listed under "Registrant" section in https://github.com/internetee/registry/issues/377#issuecomment-548913083.

https://prnt.sc/pt7n70 UI is quite poor, but it was so before. Will be addressed in a separate PR later.

### DIFF
--- a/app/views/admin/domains/partials/_dnskeys.haml
+++ b/app/views/admin/domains/partials/_dnskeys.haml
@@ -1,5 +1,5 @@
 - panel_class = @domain.errors.messages[:dnskeys] ? 'panel-danger' : 'panel-default'
-#dnskeys.panel{class: panel_class}
+.panel{class: panel_class}
   .panel-heading.clearfix
     = t(:dnskeys)
   .table-responsive
--- a/app/views/admin/domains/partials/_dnskeys.haml
+++ b/app/views/admin/domains/partials/_dnskeys.haml
@@ -1,5 +1,4 @@
-- panel_class = @domain.errors.messages[:dnskeys] ? 'panel-danger' : 'panel-default'
-.panel{class: panel_class}
+.panel.panel-default
   .panel-heading.clearfix
     = t(:dnskeys)
   .table-responsive
@@ -17,9 +16,3 @@
             %td= x.protocol
             %td= x.alg
             %td= x.public_key
-      - if @domain.errors.messages[:dnskeys]
-        %tfoot
-          - @domain.errors.messages[:dnskeys].each do |x|
-            %tr
-              %td{colspan: 4}= x
-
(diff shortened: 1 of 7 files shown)


teadur:
@artur-beljajev whats the reasoning behind hiding the errors in ui ?

artur-beljajev:
@teadur As the commit 2bdba16563d13b2eaf670acab8bd9e78a0e4d4f7 message states:

> This feature was unused.

teadur:
@artur-beljajev the pull request title is really misleading, could you improve it in a fashion that it actually shows the intent of the changes ?

artur-beljajev:
@teadur In what way? Suggest alternative.

teadur:
> @teadur In what way? Suggest alternative.

The code you are removing is used, how "remove unused code" isn't misleading ? 
What is the intention of the pull request is known by you, how can i suggest alternative for it ?

artur-beljajev:
Based on the info I got from @vohmar, it is not used.

teadur:
> Based on the info I got from @vohmar, it is not used.
So you are saying that code is never loaded henche not used ?

vohmar:
I tend to agree with @teadur on this as remove unused code or commit messages like remove unnecessary html are hardly helpful to anyone forcing one to look at the actual code change to get an idea what is being changed or removed. This has been somewhat a problem with similar changes for a long time. Isn't it better to be more specific as you actually know what is the (intended) functionality you are removing - used or unused doesn't really matter. ie "remove code to stress invalid dns and dnskey data in UI as useless" or something in those lines

artur-beljajev:
@vohmar: I am confused. @teadur talks about PR, you about commit messages. What exactly is wrong and in what way? If we take this comment as an example: https://github.com/internetee/registry/pull/1411/commits/398d59cc561dcb726128546298c5a8669146ac1a. Its message describes what has been done in general, and explains _why_ it has been removed (unused, unneeded, unnecessary). This is the intention. If you want to know _what_ exactly HTML is removed, you can see changes.

If you see a room for improvement, suggest alternative names, otherwise this discussion is too broad and I really miss the point.

vohmar:
I just extended @teadur's comment a bit sharing that I have had similar type of feelings about misleading and uninformative PRs and commit messages before. Felt like a good place to mention that as you @artur-beljajev dont seem to share the opinion raised by @teadur. I think that fixing a typo is totally different thing than removing a piece of code. Latter requiring more context at least in my opinion. You yourself should know that by how much time you put into figuring out whether it is possible or sensibel to remove something - how can you then except that others will get the whole picture by just reading vague comment like "remove unused code". Code can be unused for many reasons (because of some settings - changing some settings could make it used; its unfinished waiting the right time; it was used but is obsolete now and was forgotten to be removed; implementation is unsuccessful and users do not want to use it; etc. Furthermore every line of code is supposedly there for a reason - intended for something. So explaining in commits and in more general terms in PRs when there is no issue ticket associated with it, seems reasonable and respectful to your peers as you have meade the research to come to the conclusion that removal is the right way to go - readers have not done it so why not help them to understand the intended purpose of the code as well as reasoning behind removing it.

PR #1420, opened 2019-11-12
log error when cert_path is missing

teadur:
Un configured cert_path in epp deployment will lead to registrar portal failing to connect to epp/repp with hard to debug error "no implicit conversion of nil into String" what refers to https://github.com/internetee/registry/compare/log-error-when-cert-path-is-missing?expand=1#diff-4e75b419751f5949efd2f5b537a19d4bR19

There are so many bad practices around the whole cert handling  here i'm not even trying to fix them in this pr, rather provide meaningful error message to sysadmins to aid in debugging, this is only temporary/half solution and all exceptions here should be handled gracefully in future.

Ignore the commit name will be squash-merged


app/controllers/epp/sessions_controller.rb
```diff
@@ -13,6 +13,9 @@ def login
       webclient_request = ENV['webclient_ips'].split(',').map(&:strip).include?(request.ip)
       if webclient_request && !Rails.env.test? && !Rails.env.development?
         client_md5 = Certificate.parse_md_from_string(request.env['HTTP_SSL_CLIENT_CERT'])
+        if ENV['cert_path'].nil?
```
artur-beljajev: - `present?` is better here, since it also handles empty string, whereas `nil?` doesn't.

app/controllers/epp/sessions_controller.rb
```diff
@@ -13,6 +13,9 @@ def login
...
+          logger.error('webclient cert (cert_path) missing, registrar epp/repp disabled')
```
artur-beljajev: I don't see a point continuing execution if there is no `cert_path` configured in this case, therefore `fail 'message'` might be better here.

teadur: Could you give me example don't exactly understand what you mean

artur-beljajev: I was thinking about something like `fail 'webclient cert (cert_path) missing'` unless `ENV['cert_path']`.present?`.

app/controllers/epp/sessions_controller.rb
teadur: styleguide seems to prefer blank

app/controllers/epp/sessions_controller.rb
teadur: style guide seems to prefer raise over fail

yulgolem (approved):

Issue #1428, opened 2019-12-02
ForceDelete: forcedelete2.0 - new process


vohmar:
- [x] Statuses: currently forcedelete marks list of statuses that are set with it (serverRenewProhibited, serverTransferProhibited, serverUpdateProhibited, pendingDelete and serverManualInzone). New process sets only two - serverRenewProhibited, serverTransferProhibited
- [x] New logic - currently force delete is 30 days and started at the moment of admin setting the status. New process is minimum 45 days, but may vary to up to 1 year + 45 days. ForceDelete statuses are set immediately on starting the process, but start of the last 45 days count may be delayed. There are two versions: fast track and soft
  - [x] new parameter - force delete start date marking the beginning of the last 45 day period
  - [x] fast track version - is initiated when the signal of non existent registrant comes from a third party. New set of statuses, 45 day period starts immediately (forcedelete start date is the date forceDelete is initiated), clientHold for removing domain from the zone.
  - [x] soft version - is initiated when invalid registrant data is discovered by the registry. New set of statuses, the last 45 day grace period is postponed for up to a year, clientHold if force delete start date is different from expiration date
    - [x] formula to calculate the force delete start date is: current expiration date - full years of (current expiration date - today's date)
      - [x] if the result is 1 or more calculate and set forcedelete start date
      - [x] else no need to set the date - set statuses will regulate the usage of the domain and unless data is fixed domain registration is expired and released as per regular expiry
- [x] clientHold - new status for removing domain names from zone. client prefix means that registrar is able to remove this status in case they have contact with registrar and the process of fixing the data has begun. This status is used in the last 30 day period only if force delete start date is less than expire date. (In case the last 30 days of force delete overlap with the last 30 days of normal expiration period and the domain registrant data is not fixed and domain registration renewed, registry will automatically set serverHold at the beginning of next day)
  - [x] EPP - remove clientHold status
  - [ ] clientHold removal option to registrar portal (ui and xml console) [#1481 ]
- [x] Stopping force delete process
  - [x] Manual - Admin interface
  - [x] Automatic - registrant change - replacing current registarant object with new valid one automatically stops force delete procedure (statuses are removed, force delete start and force delete dates are removed, whois records are updated) [#1479]
- [x] Notification:
  - [x] EPP poll message on setting the force delete status (Same as with current process) [#1478]
[...]

PR #1430, opened 2019-12-03
Add new ForceDelete procedures

yulgolem:
Add new test file new_force_delete_test.rb to check if new ForceDeleteProcedures are correct.

See #1428


yulgolem:
Need to be run daily DomainCron.start_client_hold

app/models/domain_cron.rb
```diff
@@ -78,4 +78,46 @@ def self.start_redemption_grace_period
...
+    real = 0
+
+    Domain.force_delete_scheduled.each do |domain|
+      next unless domain.client_holdable?
+
+      real += 1
+      domain.statuses << DomainStatus::CLIENT_HOLD
+      log_start_client_hold(domain)
+
+      domain.save(validate: false) and marked += 1
```
karlerikounapuu: Maybe use `&&` ?

yulgolem: Well, that's some semantical && ideological point here.
You see, operators 'and' and 'or' are low precedence flow control operators useful to chain methods, not boolean operators like '&&' and '||'.
See http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/ for some examples.

karlerikounapuu: Oh, right, missed the real intention of it. Thanks for clarification

PR #1453, opened 2019-12-19
Remove revoke certificate action button from API users

### LINKED ISSUE #952: Auth: Revoking cert does not stop user from logging in
If admin revokes user's PKI cert user can still login with this certificate.
Only deleting the certificate stops user from logging in.

karlerikounapuu:
Closes #952 

As per current limitations to cert authentication design, it's easier to simply revoke & delete (instead of just revoking) registrar certificates to disable user access to registrar portal.

As this functionality is already working as expected via big red "delete" button, it renders "revoke certificate" button as useless thus can be removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+30.04.2020
+* Fix for internal error on opening domain history with legacy id record [#1576](https://github.com/internetee/registry/issues/1576)
+
 27.04.2020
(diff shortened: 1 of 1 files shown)


teadur:
@karlerikounapuu according to  #952 revoking does not work as intended.
Do i understand correctly that we drop revoke cert functionality as broken and un-needed ? 
Generally in CA world i don't think that deleting and revoking a cert is considered same action.

Issue #762, opened 2018-03-07
EPP: client side status editing


vohmar:
Attempts to set/change client side statuses over EPP fail even though Client side status editing is enabled in Admin interface (/admin/settings)

sample query:
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<epp xmlns="https://epp.tld.ee/schema/epp-ee-1.0.xsd">
  <command>
    <update>
      <domain:update
       xmlns:domain="https://epp.tld.ee/schema/domain-eis-1.0.xsd">
        <domain:name>timo-6a2.ee</domain:name>
        <domain:add>
<domain:status s="clientRenewProhibited" lang="en">Test</domain:status>
        </domain:add>
      </domain:update>
    </update>
    <clTRID>timo-1520436100</clTRID>
  </command>
</epp>
```

response:
```
<?xml version="1.0" encoding="UTF-8"?>
<epp xmlns="https://epp.tld.ee/schema/epp-ee-1.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="lib/schemas/epp-ee-1.0.xsd">
  <response>
    <result code="2306">
      <msg lang="en">Parameter value policy error. Client-side object status management not supported: status [status]</msg>
    </result>
    <trID>
      <clTRID>timo-1520436100</clTRID>
      <svTRID>ccReg-4534604104</svTRID>
    </trID>
  </response>
</epp>
```

yulgolem:
At #1454 added allowance only for `clientHold`.
Maybe we should set kind of whitelist of domain statuses we are allowing to change via EPP?

vohmar:
whitelist sounds like a good idea. We will probably remain with only hold status editing enabled for some time - so we want to allow limited client side status editing

vohmar:
registar can now add and remove clientHold statuses

PR #1454, opened 2019-12-23
Fix epp domains controller to apply add or rem changes to status

yulgolem:
From this patch epp domains controller will allow to update
domain statuses only if there is only one status - clientHold.
This was made to allow usage of new ForceDelete procedures.

See #762, #1428

### DIFF
--- a/app/controllers/epp/domains_controller.rb
+++ b/app/controllers/epp/domains_controller.rb
@@ -311,11 +311,17 @@ module Epp
 
     def status_editing_disabled
       return true if Setting.client_status_editing_enabled
+      return true if check_client_hold
       return true if params[:parsed_frame].css('status').empty?
       epp_errors << {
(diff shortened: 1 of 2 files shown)


app/controllers/epp/domains_controller.rb
```diff
@@ -311,13 +311,19 @@ def find_password
...
+      return true if check_client_hold
       return true if params[:parsed_frame].css('status').empty?
       epp_errors << {
         code: '2306',
         msg: "#{I18n.t(:client_side_status_editing_error)}: status [status]"
       }
     end
 
+    def check_client_hold
+      statuses = params[:parsed_frame].css('status').map { |frame| frame['s'] }
```
artur-beljajev: Block's parameter is called `frame`, but I guess in fact it is an XML element.

PR #1503, opened 2020-01-31
Bind specific invoice when paid via Banklink / EveryPay

### LINKED ISSUE #1500: Banklink payments fail
at the very end of banklink payment process after clicking back to merchant user is returned to registrar portal and met there by message "something went wrong". Money is removed form the account, transferred to EIS account, so transaction is successful, but invoice is not marked as paid and the credit is not loaded to registrars prepayment account.

image

there are no errors to be found in the logs. Log excerpt in chat

karlerikounapuu:
Closes #1500 

Creates new BankTransaction in case previous one is already binded with account activity. When attempting to complete transaction, it attempts to bind specific invoice directly (bind_invoice) instead of autobinding (autobind_invoices).

### DIFF
--- a/app/models/payment_orders/bank_link.rb
+++ b/app/models/payment_orders/bank_link.rb
@@ -51,11 +51,7 @@ module PaymentOrders
     def complete_transaction
       return unless valid_successful_transaction?
 
-      transaction = BankTransaction.find_by(
-        description: invoice.order,
-        currency: invoice.currency,
-        iban: invoice.seller_iban
-      )
+      transaction = compose_or_find_transaction
 
       transaction.sum             = response['VK_AMOUNT']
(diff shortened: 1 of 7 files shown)


app/models/payment_orders/bank_link.rb
```diff
@@ -57,6 +57,18 @@ def complete_transaction
         iban: invoice.seller_iban
       )
 
+      # Transaction already autobinded (possibly) invalid invoice
+      if transaction.binded?
```
yulgolem: Looks similar to [this method](https://github.com/internetee/registry/pull/1503/files#diff-e501c6a999b2d55ca6d179a0d68bbadfR40) . Maybe we may move it to, say, concern and use the same method of bank transaction creation in both cases?

Issue #1509, opened 2020-02-04
Resend e-invoices if any error occurs


yulgolem:
See https://github.com/internetee/e_invoice/issues/2, resolving on registry side.

yulgolem:
@vohmar , could you, please, describe the situations in which we are sending e-invoices from registry?

I did some additions to registrar#issue_prepayment_invoice method but not 100% sure it's the only occurrence.

yulgolem:
OK, seems what the e-invoice sending is from registrar only.

vohmar:
yes we currently send e-invoices only from registrar portal, but we should add this feature to auction as well in near future - just to keep in mind

vohmar:
closed by #1510

PR #1510, opened 2020-02-04
Add idempotent Que job with retries for e-invoice sending

yulgolem:
See #1509

### DIFF
--- a/app/jobs/send_e_invoice_job.rb
+++ b/app/jobs/send_e_invoice_job.rb
@@ -1,19 +1,17 @@
 class SendEInvoiceJob < Que::Job
-
   def run(invoice)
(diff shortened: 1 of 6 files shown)


vohmar:
getting an error on sending an invoice

`Feb 17 12:32:39 staging st-adm.infra.tld.ee[15149]: {"lib":"que","hostname":"staging","pid":15149,"thread":47145360343740,"event":"job_errored","job":{"queue":"","priority":100,"run_at":"2020-02-17 10:32:38 +0000","job_id":449168,"job_class":"SendEInvoiceJob","args":["Invoice no. 3131077"],"error_count":3},"error":{"class":"NoMethodError","message":"undefined method `id' for \"Invoice no. 3131077\":String"}}`

Issue #205, opened 2016-10-11
Directo: Multilingual support for monthly invoices


vohmar:
Add multilingual support for monthly invoices sent from Directo. There is a language parameter in the invoice XML sent to Directo (https://github.com/internetee/registry/blob/6eef658a992710d354859a32d43e1c2a0ccd497e/app/models/directo.rb)

The language parameter is currently left empty. 

We need a rule that specifies language parameter as ENG in case registrar country_code value is not EE

artur-beljajev:
@teadur or @ratM1n Check that `select count(*) from registrars where country_code is null` returns 0 on prod.

artur-beljajev:
See #509

artur-beljajev:
@vohmar Please review #640. I need that for this ticket.

yulgolem:
Added 'Language' field to the Directo. ATM sends registrar.language as is (two-letter ISO 639-1 codes). As we are using some code in a product name, don't really think we should translate that.

![image](https://user-images.githubusercontent.com/7688888/74134739-dda33280-4c0c-11ea-941a-56e0c12acf93.png)

vohmar:
tested this pr - set country for a registrar to Finland and run the monthly cron task, invoice was created and sent to Directo but no language parameter was set in xml:

<?xml version=\"1.0\" enco
ding=\"UTF-8\"?><invoices>  <invoice Number=\"309907\" InvoiceDate=\"2020-01-31\" PaymentTerm=\"R\" CustomerCode=\"RIKS\
" Language=\"\" Currency=\"EUR\" SalesAgent=\"JAANA\">    <line RN=\"1\" RR=\"1\" ProductName=\"Domeenide registreerimin
e - Jaanuar 2020\"/>    <line RN=\"2\" RR=\"2\" Quantity=\"1\" ProductID=\"01EE\" Unit=\"tk\" ProductName=\".ee registre
erimine: 1 aasta\" UnitPriceWoVAT=\"6.0\"/>    <line RN=\"3\" RR=\"3\" Quantity=\"-1\" ProductID=\"ETTEM06\" Unit=\"tk\"
 ProductName=\"Domeenide ettemaks\" UnitPriceWoVAT=\"6.0\"/>  </invoice></invoices>

yulgolem:
ATM we are checking not for country, but for registrar's language.
If it set for 'en', then language field for invoice is set for 'ENG', otherwise nil.

[...]

PR #1523, opened 2020-02-10
Add language field to the Directo model

yulgolem:
See #205


yulgolem:
> Language property should be "ENG" for english customers

And for Estonian?
Haven't found any on it yet, but in Omnivia e-invoice description was the clause of them using ISO 639-1 codes, so I assume Directo may use this approach as well. But not 100% sure.

yulgolem:
Changed English language value to ENG, and blank if Estonian.

PR #1527, opened 2020-02-12
Registrant: Allow to view other contacts of domain

karlerikounapuu:
If registrant has ties with domain, he/she can now also view contacts that are tied to the same domain.


vohmar:
none of the registrant objects open now - internal error is raised

error from browser console: GET https://st-rant.infra.tld.ee/registrant/domains/86927/contacts/238306 404 (Not Found)

Issue #1298, opened 2019-09-10
Move to consistent data_migrations


teadur:
We have been using different methods how we migrate data in our database.
1) At first data_migrate gem was used ( https://github.com/internetee/data-migrate )
That introduced db/data directory with some migrations and data_migrate table (what was removed in https://github.com/internetee/registry/commit/62ec30c708bfc3b87974aa4b6130c7f8ae76ca47 ) 
Our Gemfile still requires data_migrate.
If we want to use data_migrate gem we should move to current version of it ( https://github.com/ilyakatz/data-migrate ) 

2) We have migrated some data inside normal activerecord migrations 
https://github.com/internetee/registry/commit/38f967a7d64d586262c7e9da6eb3c66576491c9c

3) We have written some data migrations as rake tasks ( Very hard to figure out the correct ordering / at what point of time they should be applied ) 

https://github.com/internetee/registry/search?q=data_migrations&unscoped_q=data_migrations

We need to figure out how we want to migrate data and stick to one method.
I propose we upgrade data_migrate gem and use only that for data related migrations.

vohmar:
fixed by #1539

PR #1539, opened 2020-02-25
Move data migrations from rake tasks to gem

yulgolem:
See #1298


yulgolem:
Need to add `rake data:migrate` to deploy script.

PR #1570, opened 2020-04-09
Registrar: Disable CSRF validation on creating new session

karlerikounapuu:


vohmar:
waiting for confirmation from the registrar if the problem persists and wether it might be caused by the registrars own proxy server or setup for remote work

karlerikounapuu:
Not relevant anymore

PR #1574, opened 2020-04-27
Implement disputed logic for domains

### LINKED ISSUE #269: Domain disputes: special reserved list (dispute domains)
As of 1st of January 2017 the domains registrations that have been successfully disputed must be reserved for the contender for up to 3 years.

* new list (dispute domains) similar to reserved domains list (/admin/reserved_domains) - columns name, pw, created/updated at, valid to, actions. Filter fields - name, valid to from , valid to until, results per page
* domain form for adding domains to the list similar to reserved form (/admin/reserved_domains/new) with additions of mandatory 'valid to' and 'comment' (for url to dispute decision) fields
* registry must check that in case of epp domain update query with \<domain:chg>\<domain:registrant> element (ie registrant change) the domain is not in the dispute domains list - if it is in the list the \<eis:extdata>\<eis:reserved> element with correct password must be present in the extension block
* registry must check that in case of epp domain create query the domain name is not in the reserved nor dispute domains lists - if it is the \<eis:extdata>\<eis:reserved> element with correct password must be present in the extension block
* if a domain name that is in the reserved domains list is added to the dispute domains list reserved list password is replaced with the new password from dispute domains list (if a domain in the dispute list is added to the reserved list the password in the reserved list is set from the dispute list - admin needs to be notified if the optional password field had a value in it - "proposed password rejected, dispute password used instead") - password for any domain in both reserved and dispute lists must be the same (shared)
* if the password for a domain in dispute list is used for domain create or domain update the domain is deleted from the dispute list (in case of reserved domain - the reserved password is reset like in regular reserved domain registration)
* The domains in dispute list will get a dynamic disputed status in WHOIS identically to current reserved and blocked statuses. Described in sub-issues https://github.com/internetee/rest-whois/issues/29 and https://github.com/internetee/whois/issues/15
* If the contender does not register the domain within the period of 3 years since the dispute decision (valid to of dispute list record) the domain will be removed from dispute list.

---
To test:
1. WHOIS regeneration: on domain update
2. WHOIS regeneration: on contact update
3. WHOIS regeneration: manually via rake whois:regenerate
4. /registrant/whois
5. WHOIS regeneration: on domain delete
6. Dispute/Blocked Domain/Reserved Domain management

karlerikounapuu:
Closes #269 

Requires addition of `DisputeStatusUpdateJob.run` job once per day (12AM) for applying in-future dispute statuses.
 
`dispute_period_in_months: 36` must be added to `config/app.yml`, if not present

- [x] disputed status is missing - not present in registry/epp not whois
- [x] restwhois ei toimi - uurin martiniga mis seal viga võib olla
- [x] please use reserved element for disputed password - adding disputed element means development for registrars and that might be a problem
- [x] domain update / edit view needs an additional field to enter the reserved password - ideally this field should be dynamic, available only if the domain in question is disputed (consider effort - this is nice to have)
- [x] if verified yes element is missing from domain update the update request is put in bending but domain is already reamoved from the disputed list. If request is not confirmed wihting 48 hours the pending requet is canceled, previous owner remains as registrant but domain is free of disputed restrictions. Prefered solution here would be to autoconfirm reigstrant change id correct password is provided.
- [x] if admin deletes the disputed record it is not moved to expired / closed disputes list
- [x] there should not be a delete button for expired/closed disputes - as we do not have archive table for disputes (ie log_disputes) we need to save history. Instead we should show the date when the record was removed from active dipsutes list and if possible together with reason (registered/expired/deleted)
- [x] in the add to disputed view under start at cell there is a comment "Must be at least today / in future" that is not true. User can enter any date in the field. Actually the date should be today or in the past becuase domain should not be inserted to disputed list unless the dispute commitee's decision is in force. But this gives an alternative use case of not adding any manual statuses to prohibit unwanted updates to the registration. Will discuss this with registry admin.
- [x] removing a domain from disputed list while it is not registered should put the domain in auction list (if the listing expires or admin deletes it)


app/jobs/dispute_status_update_job.rb
```diff
@@ -0,0 +1,61 @@
...
+    disputes = Dispute.where(closed: false, starts_at: Time.zone.today).all
+    Rails.logger.info "DisputeStatusUpdateJob - Found #{disputes.count} activatable disputes"
+
+    disputes.each do |dispute|
+      process_dispute(dispute, closing: false)
+    end
+  end
+
+  def process_dispute(dispute, closing: false)
+    intent = closing ? 'close' : 'activate'
```
yulgolem, line 33: Do you really want to set the value of 'intent' variable or just to check it? If latter, than '==', suppose.

app/models/epp/domain.rb
```diff
@@ -475,6 +478,20 @@ def update(frame, current_user, verify = true)
...
+    if !same_registrant_as_current && disputed?
+      disputed_pw = frame.css('disputed > pw').text
+      if disputed_pw.blank?
+        add_epp_error('2304', nil, nil, 'Required parameter missing; disputed' \
+        'pw element required for dispute domains')
+      else
+        dispute = Dispute.active.find_by(domain_name: name, password: disputed_pw)
+        if dispute.nil?
+          add_epp_error('2202', nil, nil, 'Invalid authorization information; '\
+          'invalid disputed>pw value')
```
yulgolem, line 490: Suppose that's a typo, don't we need underscore here? LIke 'disputed_pw'

app/views/admin/disputes/_form.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+    <div class="row">
+        <div class="col-md-8">
+            <div class="panel panel-default">
+                <div class="panel-heading clearfix">
+                    <div class="pull-left">
+                        <%= t(:general) %>
+                    </div>
+                </div>
+                <div class="panel-body">
```
yulgolem, line 12: If next div is inside this one, maybe we could show it by indentation?
That's why I personally prefer haml or maybe slim- it's a bit easier to read, IMO.

app/jobs/dispute_status_update_job.rb
karlerikounapuu: It was supposed to set worded value for `intent`, so it can be later used for wording / manipulating named hash values.

app/models/epp/domain.rb
karlerikounapuu: Actually this was intentional as well, as same structure was used before for reserved domains / other auths as well. EPP has an object `disputed` which has a `pw` field.

app/views/admin/disputes/_form.html.erb
karlerikounapuu: Both `panel-body` and` panel-heading clearfix` divs are supposed to same-level children of `panel panel-default`

karlerikounapuu: Nvm, got what you meant. Thanks!

PR #1579, opened 2020-05-14
Simple implementation of retained domains API endpoint

maciej-szlosarczyk:
This needed a new name, there are several classes of database object to be included in that endpoint. Retains existing REPP URL schema.

Currently, there is one list ordered by name, with each object containing status and ascii name for convenience. Can be converted to multiple fields (reserved and blocked separately). Also contains total count.

Includes CORS preflight handling which seems to be a known problem for Rails in the past.

@vohmar Testing: Open URL /repp/v1/retained_domains and you should see a JSON object there.

I'll write required documentation once the questions about the shape of the data are answered, also will add then strict checking of content type.


maciej-szlosarczyk:
@vohmar I added documentation as well. Caching uses existing ETag mechanism, which is part of HTTP 1.1 standard that should be known to developers who will integrate this on registrars' side, but I still wrote it down.

For testing purposes:
When you make a request, you can nowe pass type parameter, it can be either reserved or blocked. cURL example:

```
$ curl -v "https://registry.test/repp/v1/retained_domains?type=blocked"
```

Response should look something like this:

```
< HTTP/1.1 200 OK
< Date: Fri, 15 May 2020 12:02:54 GMT
< Content-Type: application/json; charset=utf-8
< ETag: W/"d4a387cd5cc32e213f66c4a9493bef39"
< Cache-Control: max-age=0, private, must-revalidate
< X-Request-Id: dd3188c4-5c44-4bf7-901e-838db819dcfd
< X-Runtime: 1.105368
< Transfer-Encoding: chunked
```

Now you can use the ETag to make another request to check cache (please note the escape slashes in If-None-Match header):

```
$ $ curl -v -H "If-None-Match: W/\"a905b531243a6b0be42beb9d6ce60619\"" "https://registry.test/repp/v1/retained_domains?type=reserved"
```

And the response should be a 304 status with no body. If the tag is not valid (i.e the list has changed), the response will be transferred normally with status 200.

PR #1584, opened 2020-05-22
Ensure all tests run on CI

maciej-szlosarczyk:
Fixes tests that were missing environment config and such. I've re-used the test CA that I put into `epp_proxy` at some specific point.

@karlerikounapuu please ping me when you untangle the issue with contact viewing.


maciej-szlosarczyk:
There is an issue with some intermittent test failures, but should be good to go for now. If the failures persist, we'd need to investigate more.

vohmar:
has a merge conflict now that disputes branch is merged

PR #1593, opened 2020-06-01
Upgrade Rails from 5.2 to 6.0.3

karlerikounapuu:
Upgrades Rails to 6.0.3

Added bootsnap, parallel testing.

Did not implement Zeitwerk / Webpack yet, even though it's new standard from RoR 6.

What should be tested definitely:
- Static asset generation, as sass-rails 6.0 fires sassc-rails which caused errors in rest-whois
- PaperTrail, are versions saved correctly as they should, including relations.
- File upload/download, such as invoices, csvs, pdfs


config/environments/development.rb
```diff
@@ -0,0 +1,66 @@
+Rails.application.configure do
```
maciej-szlosarczyk, line 1: I missed this earlier, but this is a bit inconvenient. Development configuration should not be committed to source control, it is not the same for everyone.

We have a file called development.rb.sample where the default configuration should be put and every developer can adjust it to their local setup.

I'll make a PR to correct this.

PR #1596, opened 2020-06-03
297 email address validation truemail

### LINKED ISSUE #297: email address validation and email processing
after dropping postal addresses the importance of the remaining contact data incl email address is that much higher.

email address validations:
1. entered email address is correctly formed as email address
2. domain is registered and in zone (internal check for local zones, dns query for the domain part of email address for other TLDs)
3. there is an email server answering on that domain (dns query for mx record and ping?)

on contact:create and contact:update for email address all the validations should be run

on domain:create, domain:udpate, domain:renew and before sending any emails internal domain registration verification must be performed

all emails sent by the registry must go out using Que
email sending que jobs must be configured so that in case of fatal error - ie nxdomain , malformed e-mail address the error is logged and the que job does not get re-run. In case of potentially temporary errors - mailserver not reploying, mailbox rejecting the email there must be a limit on how many times or for how long does the registry retry to send the email ie up to 48 hours.

yulgolem:
Closes #297


app/models/registrar.rb
```diff
@@ -28,14 +29,11 @@ class Registrar < ApplicationRecord
   validates :vat_rate, numericality: { greater_than_or_equal_to: 0, less_than: 100 },
             allow_nil: true
 
-  validate :forbid_special_code
```
yulgolem, line 31: Duplicates line 24 validation.

app/models/nameserver.rb
```diff
@@ -88,7 +88,7 @@ def glue_record_required?
   end
 
   def normalize_attributes
-    self.hostname = hostname.try(:strip).try(:downcase)
+    self.hostname = hostname.try(:strip).try(:downcase).gsub(/\.$/, '')
```
yulgolem, line 91: For some reason we were asserting :strip method to remove last dot from string. It is not, so for tests to be OK and to continue usual behaviour I've added :gsub.

test/test_helper.rb
```diff
@@ -43,7 +43,7 @@ def representation_rights(citizen_personal_code:, citizen_country_code:)
 
 class ActiveSupport::TestCase
   # Run tests in parallel with specified workers
-  parallelize(workers: :number_of_processors)
+  # parallelize(workers: :number_of_processors)
```
yulgolem, line 46: For local tests work on Mac. Not only mine issue.

db/migrate/20200605100827_create_email_address_verifications.rb
```diff
@@ -0,0 +1,13 @@
+class CreateEmailAddressVerifications < ActiveRecord::Migration[6.0]
+  def change
+    create_table :email_address_verifications do |t|
+      t.string :email, null: false
```
maciej-szlosarczyk, line 4: Postgres string type is case sensitive. `maciej@email.com` and `Maciej@email.com` are different records. I'd strongly suggest using `citext` extension in postgres to avoid unexpected behaviours and saving some time on formatting and querying the data.

yulgolem: thanks, wilco

yulgolem: Maybe that's not a huge deal - we do not interested in difference between `macjei@` and `Macieji@`, as long as its a vaild email.

maciej-szlosarczyk: It is sort-of big deal in this sense that:
* You get duplicate records on different casing of the same email.
* You then have to perform the validation multiple times.

Most importantly for me, when you have to compare emails, for example for search or sign-on, you have to remember to downcase the text before continuing with the operation. Being forgetful, I always vote for some program doing a thing for me :)

app/models/registrar.rb
```diff
@@ -11,6 +12,16 @@ class Registrar < ApplicationRecord
   has_many :nameservers, through: :domains
   has_many :whois_records
   has_many :white_ips, dependent: :destroy
+  # belongs_to :email_address_verification, class_name: 'EmailAddressVerification',
```
yulgolem, line 15: leftovers, need to clean

config/initializers/truemail.rb
```diff
@@ -0,0 +1,77 @@
+require 'truemail'
+
+Truemail.configure do |config|
+  # Required parameter. Must be an existing email on behalf of which verification will be performed
+  config.verifier_email = 'info@internet.ee'
```
teadur, line 5: Maybe re-using action_mailer_default_from what is defined in application.yml would be good idea ?
Not sure what they mean by existing email, if the account has to be present i think we maybe need to create mailbox for no-reply then, at least i would not mix role mail address and "system" mail address.

config/schedule.rb
```diff
@@ -62,6 +62,10 @@
     rake 'domain:discard'
   end
 
+  every 10.minutes do
+    rake 'verify_email:all_domains'
```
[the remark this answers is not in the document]
yulgolem: On contact/registrar creation we are also creating EmailAddressVerification. These verifications are also created by existing emails/billing emails.

Once per 10 minutes we are taking these verifications what were not verified last month grouped by email domains.
Then for each domain we are taking one random email and trying to verify it asynchronously by calling VerifyEmailsJob on it. Rinse, repeat.

config/initializers/truemail.rb
yulgolem: Yes, suppose that's a good idea.

PR #1599, opened 2020-06-04
Move dev config to sample file

maciej-szlosarczyk:
There should be no impact on production environment.

NB! This will remove existing development.rb file @karlerikounapuu @yulgolem. You can restore it by running:

```bash
$ cp config/environments/development.rb.sample config/environments/development.rb
```


config/environments/development.rb.sample
```diff
@@ -5,33 +5,64 @@ Rails.application.configure do
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
+  config.autoload = :classic
```
karlerikounapuu, line 8: This can actually be removed, residue from testing when I was wondering why I couldn't turn off Zeitwerk. Real key was `config.autoloader = :classic` and it's present in application.rb anyway as of now :)

maciej-szlosarczyk:
I have also removed the assets from precompile path as it started segfaulting on travis.

PR #1601, opened 2020-06-05
Extract Xml deserializing from EPP Contact and Domain classes

maciej-szlosarczyk:
This is first of many steps in untangling Contact/Update from EPP and converting it to a more sequential processing.

Please check that you can still attach documents and set idents via EPP, it should work as previously, since the functionality is quite well tested so I'm confident it is fine :)

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -1,3 +1,5 @@
+require 'deserializers/xml/legal_document'
+
 class Contact < ApplicationRecord
   include Versions # version/contact_version.rb
   include EppErrors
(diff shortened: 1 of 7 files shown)


maciej-szlosarczyk:
Please ignore the codeclimate issues, at this moment they are false positives and can be dealt with a a later time.

app/models/contact.rb
```diff
@@ -1,3 +1,5 @@
+require 'deserializers/xml/legal_document'
```
yulgolem, line 1: Maybe we should include deserializers in config.autoload_paths so we don't need to require them? But for sure it's up to you, just wondering.

vohmar:
Did not see any issues. Is it ok to merge?

maciej-szlosarczyk:
@vohmar Yes, anything additional will be done in the next steps :)

PR #1609, opened 2020-06-11
Fix not running whois update after contact update

### LINKED ISSUE #1604: Updating contact object fails to trigger whois data update for associated domain records
When contact object update should trigger whois update job to update data of all domain objects that the contact has relations to. At hit moment whois record is not updated when contact object is updated so in case for example email address is changed registrant contact service will still send emails to previous email address.

yulgolem:
Closes #1604

```DEPRECATION WARNING: The behavior of changed? inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after save returned (e.g. the opposite of what it returns now). To maintain the current behavior, use saved_changes? instead.

DEPRECATION WARNING: The behavior of changed_attributes inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after save returned (e.g. the opposite of what it returns now). To maintain the current behavior, use saved_changes.transform_values(&:first) instead.

DEPRECATION WARNING: The behavior of attribute_changed? inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after save returned (e.g. the opposite of what it returns now). To maintain the current behavior, use saved_change_to_attribute? instead.

DEPRECATION WARNING: The behavior of attribute_was inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after save returned (e.g. the opposite of what it returns now). To maintain the current behavior, use attribute_before_last_save instead.
```

So we need to check all the using of dirty model changes checks in all the app. Also removed silencer from test config.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -502,7 +502,8 @@ class Contact < ApplicationRecord
 
   def update_related_whois_records
     # not doing anything if no real changes
-    return if changes.slice(*(self.class.column_names - ["updated_at", "created_at", "statuses", "status_notes"])).empty?
+    ignored_columns = %w[updated_at created_at statuses status_notes]
+    return if saved_changes.slice(*(self.class.column_names - ignored_columns)).empty?
 
     names = related_domain_descriptions.keys
     UpdateWhoisRecordJob.enqueue(names, 'domain') if names.present?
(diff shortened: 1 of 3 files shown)


vohmar:
tested on staging
* added new ns to a domain - whois_records table was updated
* changed email address of the registrant - whois_records table was not updated

PR #1617, opened 2020-06-26
Split contact update into XML and Business action parts

maciej-szlosarczyk:
Separates completely XML parsing from contact update. Actions::ContactUpdate class can be re-used in REPP later.

There were so many issues in this code originally that the CodeClimate count can be safely ignored.

### DIFF
--- a/app/controllers/epp/contacts_controller.rb
+++ b/app/controllers/epp/contacts_controller.rb
@@ -1,3 +1,5 @@
+require 'deserializers/xml/contact_update'
+
 module Epp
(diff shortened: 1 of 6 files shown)


maciej-szlosarczyk:
Please test if EPP contact update is still working as expected, and then we can merge.

vohmar:
Create contact and any kind of contact update fails with 2400 command failed

maciej-szlosarczyk:
@vohmar You can try again, I trusted the tests too much and made a naive typo, it is fixed now.

PR #1649, opened 2020-08-03
Set PaperTrail :whodunnit for all Registrar-related controllers

### LINKED ISSUE #1636: Objects are missing creator and updator strings
contact objects and registrant_verifications and may some more objects and records are missing creator and updator strings since probably upgrade to rails 6

karlerikounapuu:
Closes #1636


vohmar:
I see creator and updator strings in contact and domain table, but not in registrant_verifications for example

PR #1652, opened 2020-08-04
Domain transfer: Conditionally reuse contacts

### LINKED ISSUE #1651: Domain transfer fails with multiple contact objects with identical data in the same role
Domain transfer fails with command failed if there are multiple contact objects with identical data in the same role - different objects with the same details. This is probably caused by object merging functionality that does not have an issue when the objects are in different roles - registry attempts to assign the same object to the same domain and role multiple times.

karlerikounapuu:
Closes #1651 

Includes additional conditional when cloning domain contacts to another registrar. If similar Contact is found, `DomainContact` table is queried to see if there are any references of origin `contact_id`. If there are, we clone new object, if none, we use the similar, previously cloned one.


vohmar:
merging this to help the registrar out but, we need to redo this as we want to reuse objects as much as possible. So in this case here we should just avoid trying to apply the same domain contact role association more than once.

PR #1659, opened 2020-08-11
Replace rails-settings-cached with custom approach

### LINKED ISSUE #1629: Change settings storage mechanism
Currently Registry app uses Setting as per rails-settings-cached gem. While good in storing some seldom changing data like settings, this gem is bad in storage of, say, invoice numbers (see https://github.com/internetee/registry/issues/1600) and such due to non-transparent caching system.
We need to change this setting storage mechanist to something similar to system used in Auction app.

karlerikounapuu:
Closes #1629 

New changes to object:
- New settings are not **created** via `Setting.new_var=
`. These can be added via `SettingEntry.create(code: "string", value: "string, format: "string", group: "string")`, pretty much as in `auction_center`.
- Possible format values: `["string", "integer", "boolean", "float", "hash", "array"]`
- Group can be anything, but to appear on admin/settings, it has to be either `["domain_validation", "domain_expiration", "billing", "contacts", "other"]`

How to migrate to new version
- There is a `data_migration` for it (20200812093540_copy_legacy_settings_to_new_model.rb). This can be ran via `bundle exec rake db:data:migrate`. Please verify data integrity on admin/settings, strings _may_ have unnecessary quotes around, so make sure they look good. Although it doesn't break anything.
 
Heads up: `public.settings` table is now legacy. Settings are now queried from `public.setting_entries`

--
- [x] Create SettingEntry with forced types
- [x] Create migration functionality to copy settings from old model
- [x] Generate seed for SettingEntry to drop redundant config/app.yml
- [x] Create bridge to access SettingEntries from Setting model (or replace Setting altogether)
- [x] Test env setting loading (fixtures? seed.rb?)
- [x] Audit table


vohmar:
few issues I have spotted:
* "days_to_keep_business_registry_cache" is string, should be int
* "legal_document_is_mandatory" - was unchecked checkbox when at the start of testing, checked it tested if it has an effect (it did), then removed the tick while also changing Regsitry invoice contact value. As a result there is now text box in the ui and the value is blank instead of 0/false as one might expect from boolean type of field.
* "ds_data_allowed" is blank in the db - has value false in the original settings table
* "nameserver_required" is blank in the db - has value false in the original settings table
* "address_processing"  is blank in the db - has value false in the original settings table

It seems there is an issue with handling settings with boolean false values

vohmar:
looks much better - thank you!
there are now two unnecessary setting options in the admin UI, that should be removed: 
* Days to keep business registry cache
* Default language - the one with string type (there are two of such fields in settings, the other one worth saving is drop-down)

yulgolem (approved):

Issue #983, opened 2018-09-18
EPP & Registrar: enable subnets in ip whitelist


vohmar:
Registrars have been asking for option to enable access in subnet basis. Need to add support for subnets in ip whitelist. Support for ipv4 and ipv6 formats.

PR #1673, opened 2020-09-07
Protect domain/contact epp request with ip whitelist

yulgolem:


yulgolem:
@ratM1n let's check how is this working on staging.
Might be some issues with actual IP being not whitelisted, so I've added to application log output:
`Checking whitelisting of ip` and `Check result is`

karlerikounapuu (approved):

yulgolem:
Will be updated to close #983 .

karlerikounapuu (approved):

vohmar:
@ratM1n please check if everything works as expected - subnet based whitelisting

ratM1n:
Would be nice if already inserted addresses without network prefix will be extended as /32 or we will get `IPAddr::AddressFamilyError (address family must be specified)`

PR #1676, opened 2020-09-07
Process payments: Allow direct top up via bank transfer

### LINKED ISSUE #1101: Topping up credit account without an invoice
Implement option to add funds to credit account without creating an invoice first. Creating invoice prior instant payment is unnecessary. Allowing topping up credit balance using payment cards, bank link, or for example electronic wallets should only ask for the sum from the registered registrar user and in the result of successful transaction create receipt that is saved in the portal for registrars and automatically sent to registrar's accounting email and registries accounting system.

Additionally one could also support bank transfers for adding funds to credit balance. This would require automatic interface with e-bank that would automatically notify of new received payment and if reference number, description and registrar name match create a receipt (in case there is no matching invoice) and again send it to registrar's accounting email and registries accounting system.

karlerikounapuu:
Closes #1101

As new bank transactions are determined via LHV Connect and it doesn't support webhooks, we should decrease cycle time for `process_payments`, I'd suggest it to be launched after every 10 minutes or so.

Before merging, please merge [e_invoice#10](https://github.com/internetee/e_invoice/pull/10) beforehand. Then I could update this PR to reference master branch of `e_invoice` gem.


yulgolem (approved):

vohmar:
there is a rounding problem ie transaction for the sum of 7,4€ from latvian registrar (vat 21%) resulted with an invoice for the sum of 6,12 and vat 1,29 resulting in total of 7,41€ and newly generated invoice was not marked as paid. The invoice total should always match exactly the transferred sum in this case it is totally acceptable to use 6,12 (rounded from 6,115) and 1,28 (rounded from 1,284). As an alternative one can also just adjust the vat sum - the difference should never be more than 1 cent

vohmar:
test payment:
ref nr - NULL
description: test 313131 55

in the description the first nr is just some meaningless nr - invalid ref nr and does not match any existing invoice nr. The last nr - 55 is valid ref nr of a registrar that has no unpaid invoices

Result: lhv connect resulted with 0 transaction processed so no record in admin and no auto-generated invoice.

PR #1686, opened 2020-09-16
Reference newest version of e_invoice gem

karlerikounapuu:


vohmar:
Related to https://github.com/internetee/registry/pull/1684

PR #1687, opened 2020-09-17
Gather data about bounced emails via API

karlerikounapuu:
Related to [rest-whois#115](https://github.com/internetee/rest-whois/issues/115)

- Run database migrations

- Add `api_shared_key` key to `config/application.yml` with random value. This is shared key value between registry and rest-whois to authorize POST request.

It's expected that HTTP's `Authorization` header is arriving to rails and not thrown away.


config/routes.rb
```diff
@@ -298,6 +299,7 @@
     resources :delayed_jobs
     resources :epp_logs
     resources :repp_logs
+    resources :bounced_mail_addresses
```
yulgolem, line 302: How about add `only: %i[index show destroy]` so we are explicitly setting there will be no updates/creates?

db/structure.sql
```diff
@@ -1706,7 +1740,7 @@ ALTER SEQUENCE public.log_payment_orders_id_seq OWNED BY public.log_payment_orde
 
 
 --
--- Name: log_prices; Type: TABLE; Schema: public; Owner: -
+-- Name: log_prices; Type: TABLE; Schema: public; Owner: -; Tablespace:
```
yulgolem, line 1743: Don't think we need this kind of changes here.

yulgolem (approved):

config/routes.rb
karlerikounapuu: Good catch! Going to change it for sure

vohmar:
Amazon access info: https://intra.internet.ee/projects/infra/wiki/Amazon

test/models/bounced_mail_address_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    @bounced_mail.email = 'recipient@registry.test'
+    @bounced_mail.message_id = '010f0174a0c7d348-ea6e2fc1-0854-4073-b71f-5cecf9b0d0b2-000000'
+    @bounced_mail.bounce_type = 'Permanent'
+    @bounced_mail.bounce_subtype = 'General'
+    @bounced_mail.action = 'failed'
+    @bounced_mail.status = '5.1.1'
+    @bounced_mail.diagnostic =  'smtp; 550 5.1.1 user unknown'
+  end
+
+  def test_email_is_required
```
yulgolem, line 17: There is no need to test if field is required - we are set it presence to be validated and validation is already tested in rails.

PR #1702, opened 2020-10-06
REPP: Contact management

### LINKED ISSUE #1580: Registrar API contacts endpoint
Part of #1408.

karlerikounapuu:
Closes #1580 

- [x] Full EPP implementation of contacts
- [x] Drop Grape and follow Rails API logic for other REPP-related controllers


yulgolem (approved):

vohmar:
* in contact create post request term code is used to represent contact object id - response uses term id for the same thing. Confusing. Use term id in request as well (matches terminology familiar from EPP protocol)
* in contacts get response both internal id and object id id code is returned - internal id is not usable for registrar and thus can be removed and term id should be used instead of code for the object id similarly to previous point

* contact update put request: illegal or unsupported parameters are ignored, with the exception of ident that returns internal error (no implicit conversion of Symbol into Integer) and command completed successfully is returned to the user. That indicates that all the changes user asked for were made. Would it be possible to indicate in response which parameters were ignored or not processed if any?

*contact history records (log_contact) lack details - whodunnit is missing for all transform requests, destroy record is logged as changing all the parameter values to NULL resulting in admin/contact_versions record with no data (path instead of name, missing id, ident and registrar. Would it be possible to keep the last state of details for the contact in destroy history record?

PR #1709, opened 2020-10-12
Add registrant api companies endpoint

### LINKED ISSUE #1708: Registrant API: add endpoint for requesting business registry information
To enable registrant portal to display companies that were associated with the user while looking for the list of domains to return to user we need to add an option to request this information via registrant API. 

Registrant API queries this information from business registry and saves this to cache table.

Registrant API should return to the portal comapny name, business registry number and country code

Karl's draft:
image

yulgolem:
Closes #1708 
Depends on https://github.com/internetee/company_register/pull/2, merge only after gem update.

### DIFF
--- a/app/controllers/api/v1/registrant/companies_controller.rb
+++ b/app/controllers/api/v1/registrant/companies_controller.rb
@@ -1,0 +1,41 @@
+require 'serializers/registrant_api/company'
+
+module Api
+  module V1
+    module Registrant
+      class CompaniesController < ::Api::V1::Registrant::BaseController
+        def index
+          limit = params[:limit] || 200
+          offset = params[:offset] || 0
+
+          if limit.to_i > 200 || limit.to_i < 1
+            render(json: { errors: [{ limit: ['parameter is out of range'] }] },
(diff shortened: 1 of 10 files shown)


app/controllers/api/v1/registrant/companies_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+          result = error_result('offset') if offset < MIN_OFFSET
+          result ||= companies_result(limit, offset)
+
+          render result
+        end
+
+        def current_user_companies
+          current_registrant_user.companies
+        rescue CompanyRegister::NotAvailableError
+          nil
```
karlerikounapuu, line 21: Shouldn't it be an empty array? `companies_result` would fail if it was nil I think.

karlerikounapuu (approved):

app/controllers/api/v1/registrant/companies_controller.rb
yulgolem: Yep, you are right, thanks.

PR #1711, opened 2020-10-14
fix new version fields for deleted domains

### LINKED ISSUE #1710: Admin: domain delete records do not return domain name nor registrant data
deleted domains appear in domain history smth like this: 

image

looks like admin does not have access to the data after domain has been removed from the domains table.  The domain record is removed from domains table after deletion but in this case history functionality should always look for information in the history (log_domains) tables

yulgolem:
Closes #1710

### DIFF
--- a/app/views/admin/domain_versions/archive.haml
+++ b/app/views/admin/domain_versions/archive.haml
@@ -57,7 +57,7 @@
             - if version
               - attributes = only_present_fields(version, Domain)
               - domain = Domain.new(attributes)
-              - attach_existing_fields(version, domain)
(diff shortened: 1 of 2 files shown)


yulgolem:
Now that's the funny thing.
Since we are fixed versions in papertrail gem version pump, we started to save new object attribute values on destroy.
But we are destroying object, so all new object attribute values are nils, so instead of version of object one step before destroying we got... object without any attribute values.

So atm I'm just not setting the last object state if object event was `destroy`.

karlerikounapuu (approved):

Issue #1198, opened 2019-05-10
Unify/refactor domain delete procedures (auction on/off, delete/force delete)


artur-beljajev:

yulgolem:
So in order to incapsulate domain managing (at least deletion) procedures we are to use Interactor design pattern (https://mkdev.me/en/posts/a-couple-of-words-about-interactors-in-rails).

yulgolem:
Everything is done, the task might be closed now.

vohmar:
done, thank you!

PR #1717, opened 2020-10-19
Change jobs from Que to ActiveJob

### LINKED ISSUE #645: Que: jobs that send emails but fail must be expired in 48 hours
jobs that send emails but fail should not retry for more than 48 hrs. Currently system will keep retrying .

before canceling the que job, the system must also log the email address in question, date, reason of the failure, source (que, RegistrantChangeExpiredEmailJob etc) and use the same method as here: https://github.com/internetee/registry/pull/1687

yulgolem:
Related to #645


yulgolem:
Not sure if it will be possible to cancel domain deleting job, need to be checked.

karlerikounapuu (approved):

yulgolem:
@vohmar assuming it will be a pain in the neck testing this whole bunch of code, I might as well move jobs from Que to Active Job as per https://github.com/internetee/registry/issues/1198 , one job at a time/PR.

PR #1728, opened 2020-10-30
Add new mailer template for expired soft delete domains

### LINKED ISSUE #1725: New email template for expired domains in soft force delete process
When soft force delete is applied for domains that have less than a year until their valid to date only renewals are blocked and registration can be used as is until the end of the registration cycle or if invalid data is fixed then indefinitely. The email sent to registrant and admin contact when domain expires is the standard one not mentioning anything about restrictions on domain renewal. To improve user experience we should add new email template for the case when domain in soft force delete process expires.

text for the new template: https://docs.google.com/document/d/1bIlo-2q5PX2QLOg2jCSLDu47GK0Y9gjBGqLhZvzjQto/edit?usp=sharing

chapter: "Aegumisteade kustutusmenetluses (registreerija, halduskontakt)"

yulgolem:
Closes #1725

### DIFF
--- a/app/jobs/domain_expire_email_job.rb
+++ b/app/jobs/domain_expire_email_job.rb
@@ -4,6 +4,10 @@ class DomainExpireEmailJob < Que::Job
 
     return if domain.registered?
 
-    DomainExpireMailer.expired(domain: domain, registrar: domain.registrar).deliver_now
+    if domain.force_delete_scheduled?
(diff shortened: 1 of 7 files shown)


karlerikounapuu (approved):

vohmar:
the subject of sent emails is incorrect. It should be:
Domeen <domeeninimi.ee> on aegunud ning suunatud kustutusmenetlusse / Domain <domeeninimi.ee> has expired and directed into deletion process / Срок действия домена <domeeninimi.ee> истек

PR #1733, opened 2020-11-03
Admin: Mass action for soft force delete

### LINKED ISSUE #1177: ForceDelete: bulk forcedelete
add option to import list of force delete domains to start the process in bulk

admin should be able to set whether and which autoemail template to use for the bulk

karlerikounapuu:
Closes #1177 

CSV requires name in punycode and force delete reason. Reason can be either:
- ENTITY_BURIED(org/priv not valid anymore);
- INVALID_PHONE (invalid phone number);
- INVALID_EMAIL (invalid email)

CSV example
```
domain_name,delete_reason
liquidation.ee,ENTITY_BURIED
invalidphone.ee,INVALID_PHONE
invalidmail.ee,INVALID_EMAIL
```


yulgolem:
BTW we do not have a means of notify on PHONE atm, only EMAIL or ENTITY_BURIED.

karlerikounapuu:
> BTW we do not have a means of notify on PHONE atm, only EMAIL or ENTITY_BURIED.

Yup, noticed this from your PR as well. I'll remove it for now, but as I understand we have interest in including this one as well.

yulgolem:
> > BTW we do not have a means of notify on PHONE atm, only EMAIL or ENTITY_BURIED.
> 
> Yup, noticed this from your PR as well. I'll remove it for now, but as I understand we have interest in including this one as well in the future.

Yes, you are right.

app/models/concerns/domain/force_delete.rb
```diff
@@ -152,4 +153,16 @@ def allow_deletion
   def force_delete_fast_track_start_date
     Time.zone.today + Setting.expire_warning_period.days + Setting.redemption_grace_period.days
   end
+
+  def notify_parties(reason)
+    ent = notification_template if %w[ENTITY_BURIED EMAIL].include? reason
+    ent ||= reason.downcase
+    self.template_name = ent
```
yulgolem, line 160: Are you sure about that? Not `notification_template`, but `reason.downcase`?

karlerikounapuu: This was supposed to be for any template that's not private_person or legal_person, which had to be defined in given csv. As I noticed `notification_template` tries to decide on template on it's own, I think it can be dropped for now. Hope it's going to work as supposed to before we're going to scan all contacts post-create for invalid mails / phones.

vohmar:
Template for starting force delete procedure on invalid contact data: https://docs.google.com/document/d/1bIlo-2q5PX2QLOg2jCSLDu47GK0Y9gjBGqLhZvzjQto/edit?usp=sharing. Chapter: Kustutusmenetluse teade (vigased kontaktandmed)

vohmar:
* did not receive any emails with the long form of soft force delete (exp date further away than a year). In case of short process email is sent on expiration referencing to invalid contacts regardless of the reason specified in the CSV.
* the csv descritpion in the anction view does not specify that column headers are mandatory. IE: "Triggers soft force delete procedure for uploaded domain list. List must be in CSV format. First row of the CSV file must contain column headings with domain_name for the first and delete_reason for the second column. Values must be separated by comma. Domain names are expected to be in punycode format, valid reasons are listed below."
* Please rename mass actions in admin to bulk actions to have unified vocbulary (already used in registrar portal) and also bulk should be more correct term lingvistically as well.

PR #1749, opened 2020-11-18
Registrar: Update WHOIS after bulk nameserver change

### LINKED ISSUE #1739: Nameserver bulk change test and whois update
* We are missing whois update trigger from the nameserver bulk change feature and some tests.
* Also we are checking domain names from csv against puny code values - so we should add conversion step to the csv import to first convert values' to ascii Punycode format and then look for the match in registry

karlerikounapuu:
Closes #1739


vohmar:
looks all good, but values read from csv are case sensitive atm. Please make the input case insensitive or force all values to lowercase before looking for values in registry

vohmar:
few minor issues:
* new bulk change is not able to match domain name from csv if it is given in punycode - would it be possible to support both utc and puny forms of the input at the same time?
* if any of the domains having the "old hostname" already has the "new hostname" attached whole bulk change process fails with internal error due to unique constraint violation in db. This error should be handled and if possible instead of failing the whole process just fail the replacement for the specific domains and inform user about it
* the blue notification message is gone that used to be there at one point informing users about the results of the bulk change ie "nameservers changed for .... and failed for ..."

PR #1754, opened 2020-11-23
Move DomainDeleteConfirm job to interactor

### LINKED ISSUE #1753: Move DomainDeleteCongirmJob to interactor
Part of #1198

yulgolem:
Closes #1753

### DIFF
--- a/test/models/whois/record_test.rb
+++ b/test/models/whois/record_test.rb
@@ -70,6 +70,6 @@ class Whois::RecordTest < ActiveSupport::TestCase
   end
 
   def registration_deadline
-    Time.zone.now + 10.days
+    @registration_deadline ||= Time.zone.now + 10.days
   end
 end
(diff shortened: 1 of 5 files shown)


test/models/whois/record_test.rb
```diff
@@ -70,6 +70,6 @@ def test_updates_whois_record_from_auction_when_payment_received
   end
 
   def registration_deadline
-    Time.zone.now + 10.days
+    @registration_deadline ||= Time.zone.now + 10.days
```
yulgolem, line 73: Now there is the tricky part.
Unless we cache the value we might face the situation when this method called in sloooow Travis VM and there is no guarantee two method calls will be made in the very same second.

So - we'll cache the value at first usage and when comparing will use cached value instead of new call for Time.zone.now.

karlerikounapuu (approved):

PR #1774, opened 2020-12-09
REPP: Domain management

### LINKED ISSUE #1756: REPP: Domain management

karlerikounapuu:
Closes #1756 

API documentation can be found from https://internetee.github.io/repp-apidoc/

- [x] Rename admin_domain_contacts_attributes / tech_domain_contacts_attributes
- [x] Return simple domain dataset for different registrar specific domain (GET domains/:id)
- [x] Fix auth_code on domain registration
- [ ] Better thorough examples for nameservers and DNSSEC keys
- [x] Use same variable names where possible (eg registrant and registrant_id)
- [x] Domain change registrant attribute to type String from Array
- [x] Renew command sanity check (dates)
- [x] Get all domain nameservers via query
- [x] Auth-code, transfer_code, authorization_code - Single naming
- [x] Fix manipulating with auth_info
- [x] Remove fax fields


OlegPhenomenon:
https://internetee.github.io/repp-apidoc/#change-domain-registrant-auth-code

Unable to confirm the change of registrant. The change request is sent to the admin panel, but after accepting the request, nothing changes. It still need to confirm the change and the domain has the UpdatePending status

app/models/actions/base_action.rb
```diff
@@ -0,0 +1,24 @@
+module Actions
```
yulgolem, line 1: My proposal is to move all the Actions from Models namespace one level up, to their own namespace, or to the Interactions namespace to Interactions/Actions.

As far as I can see Actions have nothing to do with Models and data storage pattern, they are more of service object or interaction pattern. So from architectural point of view atm we got Interactions namespace dedicated to some atomic actions... and introducing /Model/Actions seems a bit weird to me. I know, I know, that's a legacy practice — so why shouldn't we introduce the new way of actions and sometimes in the future use just these actions to make entity changes?

yulgolem (approved):

app/controllers/repp/v1/domains/contacts_controller.rb
```diff
@@ -4,6 +4,49 @@ module Domains
...
+        param_group :contacts_apidoc
+        def destroy
+          cta('rem')
+        end
+
+        def cta(action = 'add')
+          params[:contacts].each { |c| c[:action] = action }
+          action = Actions::DomainUpdate.new(@domain, contact_create_params, false)
+
+          # rubocop:disable Style/AndOr
```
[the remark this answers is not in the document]
karlerikounapuu: I agree 100% with your proposal. It's a pain in the neck to exclude this very specific check each time.

app/models/actions/base_action.rb
karlerikounapuu: I think this conversion might be a good idea, thank you. As we had some previously made actions at that path I automatically continued without any hesitation.

OlegPhenomenon:
The documentation says that "Period for domain renew (3m, 6m, 9m, 1y ... 10y)", but options other than 1y return message {"code":2002,"message":"Domain Active price missing for operation with vertigo.test","data":{}}.

OlegPhenomenon:
If the domain expires after 11 years, then it cannot be renewed for a longer period

OlegPhenomenon:
Please edit the "Bulk nameserver change"  block in the documentation: endpoint address and data transfer method. This can confuse future users.

OlegPhenomenon:
When I try to do "Bulk domain transfer" with the same hostname, but specify a different IP address so that the domains have a different IP address for the hostname, it does not update anything. Is this expected or not?

OlegPhenomenon:
Bulk nameserver change: If the domain has the serverUpdateProhibited status, then the hostname exchange is successful. Is this expected?

OlegPhenomenon:
If I want to make Bulk name server change with several domains, but I specify only one domain in the array, like this:

```
{
    "data": {
        "type": "nameserver",
        "id": "ns2.ams1.gg",
        "domains": ["33.test"],
        "attributes": {
            "hostname": "ns2.ams21.gg",
            "ipv4": "127.0.32.21",
            "ipv6": ["2620:119:352::36"]
        }
    }
}
```

Then, only the hostname and ipv6 will be updated, but the ipv4 field will become empty

OlegPhenomenon:
When I do Bulk tech contact replace, the technical contacts in the domain with the serverUpdateProhibited status also change. It doesn't in skip list

OlegPhenomenon:
When I do Bulk tech contact replace, the technical contacts in the domain with the serverTECHChangeProhibited status have been changed

OlegPhenomenon:
Same problem as last time with "Add clientHold to domain". If you make a request through postman, then it writes the following:

```
{
    "code": 2003,
    "message": "Invalid parameter 'status' value {\"transfer\"=>{\"transfer_code\"=>\"51d1654af305bafcdd7ba7ee7c19290e\"}}: Must be a String"
}
```

Everything works as it should through the shell. I think this is most likely a problem in the headers. Therefore, I would indicate in the documentation what headers you use for this request

OlegPhenomenon:
Renew a specific domain

Request:
```
{
    "renew": {
        "period": 1,
        "period_unit": "y"
    }
}
```
Endpoint: https://registry.test/repp/v1/domains/hhhher.test/renew

Return an error: undefined method `[]' for nil:NilClass
Screenshot from 2021-02-11 10-14-46

In shell same issue

OlegPhenomenon:
"Get all contacts" section

Change endpoint in docs from /repp/v1/registrar/contacts to https://registry.test/repp/v1/contacts

OlegPhenomenon:
"Get a specific poll message" section

In docs change PUT method to GET in endpoint

OlegPhenomenon (approved):

vohmar:
* Get: requesting info on a domain belonging to another registrar returns 2303 - object does not exist. That is not correct. Ideally repp should return the same set of data available in public whois, but to simplify things at this point the response could also be  2305    "Object association prohibits operation" and refer to whois for more information.
* Post:
  * error in doc (very nice doc by the way - thank you!) - payload parameters description has parameters admin_domain_contact_attributes but the shell example has admin_domain_contacts_attributes (same goes for tech contact attributes)
  * how does admin_contacts_attributes and tech_contacts_attributes or even admin_contacts and tech_contacts. The domain context should be clear from the request and as expected are only contact object ids - no other attributes. Or is there some technical reason for having long variable names like admin_domains_contacts_attributes? 
  * the doc would also be a bit more helpful with more complete examples with nameservers and subarrays and dnskeys (I can provide an example or two).


... (to be continued)


Sidenote - few contact endpoint issues:
* setting and updating auth_info does not work - should it?
* lets loose fax from REPP

karlerikounapuu:
> * setting auth code does not seem to work on domain create, tried with auth_info like in change domain description and auhtorization_code like in domain_info  (GET) response.

All references to EPP transfer_code auth_info / transfer_info / authorization_code is not referenced as `transfer_code`. Reserved password for domain creation is names as `reserved_pw`

> * why is the registrant attribute array in domain change - each domain can have only one registrant at a time - so simple string type should work fine here (always replacing one object with another.

Registrant is an object in domain change because registrar needs to have an ability to submit `verified` flag

> * Why is post used throughout with domain operations? in case of contacts the post was used for create and put for update - why is domain different (post vs put)?

Mistake in doc, fixed. PUT is meant for object update, while POST is meant for creating an object, in most cases.

> * Any reason by public_key is better name for the element here - best practice for json naming?

Correct. I'm trying to maintain the same (snake_case) key naming approach in the whole REPP protocol as it's default in Rails.

> * would it be possible to add nameserver, dnssec and status manipulation also in the domain change request?

It's possible, sure. I omitted it from the domainChange intentionally to keep down the size of payloads and make the design (and endpoints) as semantic as possible.

PR #1810, opened 2021-01-22
Contact test for new registrant after domain transfer

### LINKED ISSUE #1808: Transfer fails to create new admin contact in case of shared contacts
In case of domain transfer to another registrar new contact objects are created for the new registrar to give new registrar instant access to registrant data and possibly reduce the need to create new contacts. .ee registry system reuses existing objects in transfer whenever possible.
1. Registry system checks if there is an existing contact object with identical data already present for the new registrar and uses that if one is found
2. If there is no such object new object is automatically created
3. in case objects have been reused by the previous registrar the freshly created object should now be reused for other roles as well

third point is where the process fails. 
* In case admin and tech are shared - replacing the object for the admin contact fails
* in case registrant, admin and tech are shared - replacing tech object fails (same object is successfully used for registrant and admin)
* everything works as expected in case registrant and admin is shared
* everything work as expected in case registrant and tech is shared

so to simplify the issue is there in case admin and tech objects are shared  probably because in this case the relation is described in the same domain_contact table.

OlegPhenomenon:
Closes #1808

### DIFF
--- a/test/integration/epp/domain/transfer/request_test.rb
+++ b/test/integration/epp/domain/transfer/request_test.rb
@@ -3,6 +3,7 @@ require 'test_helper'
 class EppDomainTransferRequestTest < EppTestCase
   def setup
     @domain = domains(:shop)
+    @domain_library = domains(:library)
     @new_registrar = registrars(:goodnames)
     @original_transfer_wait_time = Setting.transfer_wait_time
     Setting.transfer_wait_time = 0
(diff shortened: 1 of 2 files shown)


test/integration/epp/domain/transfer/request_test.rb
```diff
@@ -12,6 +13,105 @@ def teardown
     Setting.transfer_wait_time = @original_transfer_wait_time
   end
 
+  def test_transfer_domain_with_contacts_if_registrant_and_tech_are_shared
+    @domain.tech_domain_contacts[0].update!(contact_id: @domain.registrant.id)
+
+    # ????????
```
yulgolem, line 19: @OlegPhenomenon let's keep code clear from this kind of a comments.

yulgolem (changes requested):

karlerikounapuu (approved):

yulgolem (approved):

PR #1812, opened 2021-01-22
Add button to download filtered domains CSV

### LINKED ISSUE #1775: Admin: csv dowload option for domain list
Admin needs an option to download filtered out list of domains - similarly to the option in registrar portal. So admin for example filters out list of domains belongig to certain registrar with specific contact in the role of registrant and then wants to download the list to csv to present the list to the registrar.

yulgolem:
Closes #1775

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,6 +7,9 @@ prepare:
 plugins:
   brakeman:
     enabled: true
+    checks:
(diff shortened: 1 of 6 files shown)


vohmar:
only the 1st page on the domain list view is exported to csv, but I guess this could be seen as a feature :)

PR #1814, opened 2021-01-22
Bulk NS change: Verify CSV integrity

### LINKED ISSUE #1813: Registrar: bulk ns change - domain list is ignored if csv is missing header row
If uploaded csv for nameserver bulk change is incorrectly formatted it is ignored and the nameserver change is run for the full list of that registrar's domains.

This creates error prone situation where change potentially affects domains that it is not supposed to.

Instead if incorrectly formatted file is uploaded the whole request should fail or it should be treated as empty list - the change request must not affect any domains.

karlerikounapuu:
Closes #1813


app/controllers/registrar/domain_transfers_controller.rb
```diff
@@ -24,8 +24,8 @@ def create
         request.basic_auth(current_registrar_user.username,
                            current_registrar_user.plain_text_password)
 
-
-        response = do_request(request, uri)
+        action = Actions::DoRequest.new(request, uri)
+        response = action.call
```
yulgolem, line 28: BTW what will happen if request somehow fails?

Don't see any rescues or error processing, suppose we might add some.

yulgolem (changes requested):
From architecture point of view LGTM, but let's add some error processing - either on Actions::DoRequest side or on client side.
And tests to it as well.

test/interactions/do_request_test.rb
```diff
@@ -0,0 +1,42 @@
...
+      }).
+    to_return(status: 200, body: ["shop.test"], headers: {})
+
+    action = Actions::DoRequest.new(@request, @uri)
+    response = action.call
+
+    assert_equal response.body, ["shop.test"]
+    assert_equal response.code, "200"
+  end
+end
```
yulgolem, line 42: Please, add the empty line in the end of file.

yulgolem (approved):

PR #1817, opened 2021-01-26
Domain admin contacts bulk change feature

### LINKED ISSUE #1764: Registrar: bulk admin change
Registrar requested a feature to change admin contacts on multiple domains

This can only be done if ident data of old and new contact is identical otherwise this is considered as contact change and requires authorization from every registrant involved.

* Identically to bulk tech chance replace existing contact_id with another
* Admin contact tab should be added to the bulk change section (/registrar/bulk_change/new)
* Contact objects are replaced only if ident data of old and new object match (ident, ident type and ident country code)

OlegPhenomenon:
Closes #1764

### DIFF
--- a/test/integration/api/domain_admin_contacts_test.rb
+++ b/test/integration/api/domain_admin_contacts_test.rb
@@ -5,48 +5,48 @@ class APIDomainAdminContactsTest < ApplicationIntegrationTest
     @admin_current = domains(:shop).admin_contacts.find_by(code: 'jane-001')
     @admin_new = contacts(:william)
 
-    @admin_new.update(ident: @admin_current.ident, 
-                      ident_type: @admin_current.ident_type, 
+    @admin_new.update(ident: @admin_current.ident,
+                      ident_type: @admin_current.ident_type,
                       ident_country_code: @admin_current.ident_country_code)
   end
 
   def test_replace_all_admin_contacts_when_ident_data_doesnt_match
     @admin_new.update(ident: '777' ,
-                      ident_type: 'priv', 
+                      ident_type: 'priv',
                       ident_country_code: 'LV')
 
-    patch '/repp/v1/domains/admin_contacts', params: { current_contact_id: @admin_current,
-                                                 new_contact_id: @admin_new },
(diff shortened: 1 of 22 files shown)


test/integration/api/domain_admin_contacts_test.rb
```diff
@@ -0,0 +1,133 @@
+require 'test_helper'
+
+class APIDomainAdminContactsTest < ApplicationIntegrationTest
+  def test_replace_all_admin_contacts_of_the_current_registrar
+    patch '/repp/v1/domains/admin_contacts', params: { current_contact_id: 'william-001',
+                                                 new_contact_id: 'john-001' },
+          headers: { 'HTTP_AUTHORIZATION' => http_auth_key }
+
+    assert_nil domains(:shop).admin_contacts.find_by(code: 'william-001')
```
yulgolem, line 9: And here we are facing the problem.
At the start of the method we do not have contact with current_contact_id `william-001` as an admin contact of domain `:shop`.
image
So the whole procedure of replacing contacts looses sense since we are replacing wrong contacts.

So my proposal is the following:
1) At the start of any test method let's make sure the start data and conditions are correct (by using breakpoints like `binding.pry` and checking data like `domains(:shop).admin_contacts` and such).
2) We do need as well the test to checking if `Contact objects are replaced only if ident data of old and new object match (ident, ident type and ident country code)`, I cannot find any ATM. Let's add this.

yulgolem (changes requested):

karlerikounapuu (approved):

yulgolem (approved):

vohmar:
getting 

{
  "code": 2304,
  "message": "Admin contacts must be identical",
  "data": {
  }
}

response to the change requests

yulgolem:
> getting
> 
> {
> "code": 2304,
> "message": "Admin contacts must be identical",
> "data": {
> }
> }
> 
> response to the change requests

Yep, because `Contact objects are replaced only if ident data of old and new object match (ident, ident type and ident country code)`. Or did I get something wrong?

PR #1827, opened 2021-01-28
Resolve issue with possible freezes on nil raw_frame

### LINKED ISSUE #1825: Ruby 2.7.2 got an error obfuscating empty fields
Got this kind of en error: https://github.com/internetee/epp_proxy/pull/44#issuecomment-768291203

> the log output for the error that brought this issue to daylight transformed after the latest ruby update. New error:
> 
> FrozenError (can't modify frozen String: ""):
> config/initializers/filter_parameter_logging.rb:5:in gsub!' config/initializers/filter_parameter_logging.rb:5:in block (2 levels) in
> 
> '
> further analysis revealed that this is caused by an & sign in the authifo password for domain transfer.
> Eroor 404 is returned to user, the request it self get past web server and proxy and the error seems to be returned from the application.
> 
> We should handle & and maybe some other special characters so that the whole query does not break on processing, if possible.

yulgolem:
Closes #1825

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,9 +1,9 @@
 Rails.application.configure do
   config.filter_parameters += [:password, /^frame$/, /^nokogiri_frame$/, /^parsed_frame$/]
   config.filter_parameters << lambda do |key, value|
-    if key == 'raw_frame'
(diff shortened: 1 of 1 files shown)


yulgolem:
Need to check if that'll resolve an issue with epp_proxy.

PR #1831, opened 2021-02-02
1818 bulk change when client has status update prohibited

### LINKED ISSUE #1818: Bulk change when domain has status "*UpdateProhibited"
client/serverUpdateProhibited denies any changes to domain registration done using domain:update EPP transform request. With addition of Restful EPP (REPP) with its extended functionality we have to consider the statuses for all the REPP requests that result with similar changes to the domain registration as in case of domain:update in EPP.

Changes available with domain:update request:
* replacing registrant contact
* changing authInfo
* adding and removing admin and technical contacts
* adding and removing ns (nameserver/host) objects
* adding and removing statuses
* adding and removing secDNS objects (DNSSEC keys and DSs)
* changing secDNS:maxSigLife (unimplemented in .ee - 2102 error MUST be returned if unimplemented https://tools.ietf.org/html/rfc5910#section-5.2.5)

in addition to REPP equivalent to EPP's domain:update the status should also be taken into account with bulk change operations that have same effect to domain registrations as listed above for domain:update ie these operations should also not be allowed when client/serverUpdateProhibited status is set on a domain.

OlegPhenomenon:
Closes #1818


karlerikounapuu (approved):

yulgolem:
Tests from here are added to https://github.com/internetee/registry/pull/1846, closing this PR.

PR #1834, opened 2021-02-04
IP Whitelist: Fix IPv4/IPv6 parsing for empty string

### LINKED ISSUE #1833: Adding only IPv6 address to registrar ip whitelist will result in empty string for IPv4 address
empty sting in ipv4 address list will result with inability to sign in nor sign out from the portal with internal error.

Logs: IPAddr::AddressFamilyError (address family mismatch)

The IPv4 or IPv6 value for any record if missing must be NULL.

OlegPhenomenon:
Closes #1833

### DIFF
--- a/test/models/white_ip_test.rb
+++ b/test/models/white_ip_test.rb
@@ -38,6 +38,20 @@ class WhiteIpTest < ActiveSupport::TestCase
     assert white_ip.valid?
   end
 
+  def test_validates_include_empty_ipv4
+    white_ip = WhiteIp.new
+
+    white_ip.ipv4 = ''
+    white_ip.ipv6 = '001:0db8:85a3:0000:0000:8a2e:0370:7334'
+    white_ip.registrar = registrars(:bestnames)
+
+    assert_nothing_raised { white_ip.save }
+    assert white_ip.valid?
+    
+    assert WhiteIp.include_ip?(white_ip.ipv6)
+    assert_not WhiteIp.include_ip?('192.168.1.1')
+  end
+
   private
 
   def valid_white_ip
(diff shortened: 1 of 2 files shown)


test/models/white_ip_test.rb
```diff
@@ -38,6 +38,16 @@ def test_validates_ipv6_format
     assert white_ip.valid?
   end
 
+  def test_validates_include_empty_ipv4
+    white_ip = valid_white_ip
+
+    white_ip.update(ipv4: '')
```
yulgolem, line 44: Not sure 100%, but as far as I remember `#update` method may bypass model validations.
I'd suggest to:
1. Create WhiteIp model (`white_ip = WhiteIp.new`)
2. assign `white_ip.ipv6`.
3. Assert no exception risen on saving `white_ip`
4. assert `white_ip.valid?`

yulgolem:
BTW maybe we should add a system test for that?
With filling fields, pressing buttons and such?

yulgolem (approved):

yulgolem (approved):

vohmar:
The empty ipv4 column is still filled with empty string ('') - no improvement

testing steps:
1. added new whitelisted ip record to a registrar (eedirect) in admin (/admin/registrars/2/white_ips/new) - no ipv4, ipv6 2001:7d0:83b6:5f84::27f
image
2. tried to access registrar portal - internal error
3. select	ipv4
  from	white_ips
  where	ipv6 = '2001:7d0:83b6:5f84::27f';

result - empty string
4. update	white_ips
  set	ipv4 = NULL
  where ipv6 = '2001:7d0:83b6:5f84::27f';

all works fine

yulgolem:
Suppose the solution need to be not on the saving/validating side, but on the WhiteIp checking side.
I'll take it.

PR #1835, opened 2021-02-04
EPP: Find contact without registrar code prefix

### LINKED ISSUE #1832: Registrar prefix on contact info and check requests
On contact create registrar specific prefix is added to any created contact id unless already present in the create request.

For example a registrar with code TEST will get TEST: prefix in front of any contact created ie creating contact with id 123 will results with an object that has id of TEST:123. This is also reported in the response message and is so to provide uniqueness of the codes.

At the same time on queries the code entered by the registrar is used as is. So if for example the same registrar would look for an object with id 123 the response would be not found creating confusion as create request will end with object already exists.

To fix this we should add the prefix check to these kind of queries as well if no valid prefix is already present. The difference here to keep in mind is that registrar could also look for objects belonging to other registrars so the prefix check should be done against the pool of existing registrar codes. Also there are two forms of prefixes in the system for each registrar the code itself and the old format from fred with CID pre-prefix so in this example both are ok TEST:123 and CID:TEST:123

OlegPhenomenon:
Closes #1832


yulgolem (approved):

vohmar:
works with contact:info, but not with contact:check

PR #1863, opened 2021-02-25
Automatic soft ForceDelete procedure

### LINKED ISSUE #1861: automated ForceDelete
Registry supports currently only manual setting of forceDelete in admin. Email validations, bounced email catching, personal and business registry identifier verification adds necessity for automated forceDelete. Part of forceDelete is also how it is removed when the basis for setting it is fixed (registrant/contact changed, email address updated etc). 

example:
If forcedelete is set due to bounced email from registrant the contact object becomes invalid and all the domains it is associated with, should be set to forceDelete. when the contact object is updated with valid email address the forceDelete must be cancelled for all the domains that are now valid.

Force-delete is applied only in case of invalid contacts:
Reasons for auto-forceDelete:
* bounced email (hard bounce?) - email template for invalid contact data
* invalid email (validator - currently on dns and mx level, smtp level will be added later) - email template for invalid contact data
* invalid ident (org does not exist or is liquidated; personal ident does not exist or the person is deceased) - email templates for non-existent org or priv contacts

yulgolem:
Closes #1861

### DIFF
--- a/test/models/domain/force_delete_test.rb
+++ b/test/models/domain/force_delete_test.rb
@@ -5,6 +5,11 @@ class ForceDeleteTest < ActionMailer::TestCase
     @domain = domains(:shop)
     Setting.redemption_grace_period = 30
     ActionMailer::Base.deliveries.clear
+    @old_validation_type = Truemail.configure.default_validation_type
+  end
+
+  teardown do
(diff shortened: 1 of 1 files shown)


yulgolem:
Closing the PR since development is made under #1874

PR #1871, opened 2021-03-05
Change Que for Sidekiq as job backend

### LINKED ISSUE #645: Que: jobs that send emails but fail must be expired in 48 hours

yulgolem:
Closes #645 

Jobs workbench is accessible after logging in as administrator on `/admin/sidekiq`.
~~After merging https://github.com/internetee/registry/pull/1873 it may be possible to increase concurrency in `config/sidekiq.yml`~~

Possible sidekiq systemd service example https://github.com/mperham/sidekiq/blob/master/examples/systemd/sidekiq.service
Run sidekiq manually: `bundle exec sidekiq`


ratM1n:
ok, deployment script is ready.
1. admin portal still has que admin interface instead sidekiq
2. what will we do with old qued jobs?

vohmar:
2. if possible I would leave the que running for at least 2 days after we have switched to sidekiq, so the expiration emails, and confirmation cancellations would finish as expected but all new scheduled tasks would go to sidekiq meanwhile. Anything left hanging after 2 days is most probably zombie task anyway - cannot remember if we had any tasks that scheduled run time would be more than 48hr from the time of listing

vohmar:
in parallel setup Sidekiq tasks worked fine but domain expiration email job in que ended with and error

image

yulgolem:
Command `rake data:migrate` should be run after branch deploy.

PR #1873, opened 2021-03-05
Fix namespaces for Zeitwerk class loader

### LINKED ISSUE #1872: Allow use of zeitwerk autoloader.
Since introducing Rails 6 there are some concurrency issues with standard autoloader, but we got code loading setup in application.rb preventing us from using zeitwerk.

We should fix this setup and migrate to zeitwerk in order to have multithreaded concurrency on delayed jobs.

yulgolem:
Closes #1872


karlerikounapuu (approved):

yulgolem:
Can be checked on staging as `RAILS_ENV=staging bundle exec rake zeitwerk:check`

PR #1878, opened 2021-03-11
Fix old classnames saved in notification, add tests

yulgolem:


ratM1n:
Yes, error is gone with this branch installed.

PR #1894, opened 2021-03-29
Update mimemagic version since older versions was dropped by maintainer

yulgolem:
Mimemagic gem of versions less than 0.3.7 doesn't exist anymore due to licensing issues, maintainer issued version 0.3.7.
We should update gems ASAP for registry images to be built.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 gem 'active_interaction', '~> 3.8'
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'iso8601', '0.12.1' # for dates and times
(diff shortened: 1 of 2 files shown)


vohmar:
not sure what is going on with the license, it was onverted to gpl and then back to MIT. Lets tray the latest 0.3.10 or 0.4.3 versions - there its certain that its back to MIT license https://github.com/mimemagicrb/mimemagic/commit/cd65290a28e58974ff0c5fa1d3f2ce858035f177

yulgolem:
Upped version to 0.3.10 in order to not to have high version nokogiri dependency.

yulgolem:
Closing since mimemagic 0.3.10 is already in master after csync update

PR #1902, opened 2021-04-05
implement domain statuses restore after domain locked, added test

### LINKED ISSUE #1900: Removing registry lock should not remove statuses set prior to setting it
Registry lock sets 3 restrictive statuses to a domain - serverUpdateProhibited, serverTransferProhibited, serverDeleteProhibited. Unlocking a domain removes those same 3 statuses even if some of them were set on a domain prior setting the registry lock thus creating a method for registrants to remove administratively set restrictions.

Removing registry lock should affect only those statuses that were actually set on locking the domain

there is one potentially tricky scenario to handle here. Registry Lock is effective only in the presence of all three restrictive statuses. If any of them is removed by admin the domain is not considered locked any more and registrant is once again able to lock it. But in this case statuses to be removed on unlocking are the statuses set by last and prior locking actions.

OlegPhenomenon:
Close #1900


app/models/concerns/domain/registry_lockable.rb
```diff
@@ -1,6 +1,8 @@
 module Domain::RegistryLockable
   extend ActiveSupport::Concern
 
+  # status_notes public.hstore,
```
yulgolem, line 4: That's not needed here

app/models/concerns/domain/registry_lockable.rb
```diff
@@ -9,6 +11,12 @@ def apply_registry_lock
     return unless registry_lockable?
     return if locked_by_registrant?
 
+    self.locked_domain_statuses_history = self.statuses.map do |status|
+      if LOCK_STATUSES.include? status
```
yulgolem, line 15: How about single-liner here?
Like `status if LOCK_STATUSES.include? status`

app/models/concerns/domain/registry_lockable.rb
```diff
@@ -36,7 +44,9 @@ def remove_registry_lock
 
     transaction do
       LOCK_STATUSES.each do |domain_status|
-        statuses.delete(domain_status)
+        unless self.locked_domain_statuses_history.include? domain_status
```
yulgolem, line 47: and another single-liner here

vohmar:
Issue
1. set registry lock in the registrant portal
2. remove one of the lock statuses in admin ie updateProhibited
3. reset lock in the registrant portal
4. unlock the domain in the registrant portal
problem: only the single status set on the last registry lock reset was removed - others should be removed as well since these were also added by the locking action

yulgolem:
PR was reverted due to high amount of DB load on data migrations.
Set on hold until Sidekiq will be online in prod so we could implement the PR in question in two parts:
1) Preparing the DB via async jobs
1) Updating code itself.

PR #1903, opened 2021-04-06
Remove ForceDelete if all domain emails are correct

### LINKED ISSUE #1898: Remove force delete if bounced email address is replaced
If forcedelete is set to a domain because of invalid email address (email added to bounce list or as a result of email address validator action) the status should be automatically removed once the email address for the affected contact object is changed or the affected contacts are removed from the domain.

yulgolem:
Closes #1898

### DIFF
--- a/app/interactions/domains/force_delete_lift/base.rb
+++ b/app/interactions/domains/force_delete_lift/base.rb
@@ -1,17 +1,14 @@
 module Domains
   module ForceDeleteLift
     class Base < ActiveInteraction::Base
-      string :email,
-             description: 'Email to check if ForceDelete needs to be lifted'
+      object :domain,
+             class: Domain,
+             description: 'Domain to check if ForceDelete needs to be listed'
 
       def execute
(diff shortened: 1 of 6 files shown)


yulgolem:
Delayed Que job `CheckForceDeleteLift` must be scheduled to be performed on cron once in a while. It takes all the soft-force-deleted domains with invalid emails, iterates through them and checks for each domain if contact/registrant domains are ok, and if so - cancels ForceDelete. Tests are ready as well.

OlegPhenomenon:
The force delete status is removed even if the invalid email is still in the bounce list, after CheckForceDeleteLift.perform_now

Steps:
Add email to bounce list
On the domains that refer to the email contact added force delete status
Run CheckForceDeleteLift.perform_now

vohmar:
something is indeed nor right here. It looks like the job removes force delete from all the domains that were set force delete because of the invalid email regardless if the issue has been fixed or not. 

email in bounce list
image

associated domains with a contact with that email - domains have FD set
image

had 15 FD domains, after running the job only 5 remain
image

PR #1914, opened 2021-04-09
Add check for bounce presence in #failed? method of verification

### LINKED ISSUE #1837: Invalidate emails in case of existing record in bounce list
There are two tables in the data base email_address_verifications and bounced_mail_addresses. First one holds results from the registry email address validator the other one collects addresses that have bounced back sent emails. There is no relation, so nonexistent email address with valid domain name and configured mx record in nameservers will at the moment result in verified email (green in the ui) but with record on bounce list due to non existing mailbox for that particular address.

UI in admin and perhaps in other portals should display bounced emails in red like in case of invalid addresses in verification list.

yulgolem:
Closes #1837

### DIFF
--- a/app/models/email_address_verification.rb
+++ b/app/models/email_address_verification.rb
@@ -34,11 +34,15 @@ class EmailAddressVerification < ApplicationRecord
   end
 
   def failed?
-    verified_at.present? && !success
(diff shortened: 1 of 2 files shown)


yulgolem:
This contact got email verified, but got bounce on this email address
image

PR #1936, opened 2021-04-19
implement get all list notifications repp endpoint

OlegPhenomenon:


OlegPhenomenon:
the complete list of poll messages is accessed via `https://registry.test/repp/v1/registrar/notifications/all_notifications`. Options are also available `:limit `and `:offset`

vohmar:
the returned list is not complete. Tested on staging - regsitrar has 568 unread messages, the all_notifications list returned last 200.

PR #1952, opened 2021-04-26
added errors handler

OlegPhenomenon:


app/controllers/epp/base_controller.rb
```diff
@@ -8,7 +8,7 @@ class AuthorizationError < StandardError; end
     before_action :ensure_session_id_passed
     before_action :generate_svtrid
     before_action :latin_only
-    before_action :validate_against_schema
+    # before_action :validate_against_schema
```
yulgolem, line 11: Please, remove this comment here and put this action back online.

Use skip_before_action in errors controller instead.

yulgolem (changes requested):

yulgolem (approved):

PR #1969, opened 2021-05-07
added init which returns definitly count of domains

OlegPhenomenon:
Close #[https://github.com/internetee/registrant_center/issues/54](url)

### DIFF
--- a/app/controllers/api/v1/registrant/domains_controller.rb
+++ b/app/controllers/api/v1/registrant/domains_controller.rb
@@ -6,10 +6,12 @@ module Api
       class DomainsController < ::Api::V1::Registrant::BaseController
         before_action :set_tech_flag, only: [:show]
 
+        LIMIT_DOMAIN_TOTAL = 3000.freeze
+
         def index
           limit = params[:limit] || 200
           offset = params[:offset] || 0
-          simple = params[:simple] == 'true' || false
+          simple = params[:simple] == 'true' || false         
 
           if limit.to_i > 200 || limit.to_i < 1
             render(json: { errors: [{ limit: ['parameter is out of range'] }] },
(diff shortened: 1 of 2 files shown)


OlegPhenomenon:
Related to -> https://github.com/internetee/registrant_center/issues/54

app/controllers/api/v1/registrant/domains_controller.rb
```diff
@@ -57,10 +59,34 @@ def current_user_domains_total_count
         end
 
         def current_user_domains
-          current_registrant_user.domains(admin: params[:tech] != 'true')
+          initialization_count_of__domains
         rescue CompanyRegister::NotAvailableError
+          initialization_count_of__direct_domains
+        end
+
+        def initialization_count_of__direct_domains
```
yulgolem, line 67: Double underscore?

app/controllers/api/v1/registrant/domains_controller.rb
```diff
@@ -57,10 +59,34 @@ def current_user_domains_total_count
...
+              return current_registrant_user.direct_domains(admin: false)
+            end
+
+            current_registrant_user.direct_domains(admin: true)
+          end
+
           current_registrant_user.direct_domains(admin: params[:tech] != 'true')
         end
+
+        def initialization_count_of__domains
```
yulgolem, line 79: Here as well

yulgolem (changes requested):
In Rails way "initializer" is a config file for some external lib/gem/service put into `/config/initializers` folder, so I'd ask to consider renaming the PR and entities.

BTW why shouldn't we call method just something like 'init_count_*'?

yulgolem (approved):

PR #1977, opened 2021-05-14
added feature permission for release if domain has status server-rele…

OlegPhenomenon:
…ase-prohibited


app/models/concerns/domain/releasable.rb
```diff
@@ -17,14 +17,14 @@ def releasable_domains
           ' AND ? != ALL(coalesce(statuses, array[]::varchar[]))',
               Time.zone.today,
               Time.zone.today,
-              DomainStatus::SERVER_DELETE_PROHIBITED)
```
yulgolem, line 20: We 100% sure that will be not a deleteProhibited, but releaseProhibited?

yulgolem (approved):

vohmar:
run domain:delete against domain with serverReleaseProhibited - ended up witht a domain that had a lot of unrelated statuses set: (forcedelete, renewprohibited, transferprohibited). Looks like forcedelete was set together with the delete request
image
the same does not happen without releseProhibited set or if the confirmation step is in between

vohmar:
the issue was not related to this ticket, but was expected result as the domain had registrant with bounced email address.

spotted one minor issue - in admin the new releaseProhibited status is listed as ServerReleaseProhibited whilst others are all listed without the server prefix as admin cannot add client prefixed statuses.

image

PR #2007, opened 2021-05-26
2004 update domain eis schema with server release prohibited status

### LINKED ISSUE #2004: Update domain-eis schema with serverReleaseProhibited status
There is a new status in registry serverReleaseProhibited that also needs to be added to schema file

Related to https://github.com/internetee/registry/issues/1885

OlegPhenomenon:
Close #2004


test/test_helper.rb
```diff
@@ -69,4 +68,13 @@ class ApplicationIntegrationTest < ActionDispatch::IntegrationTest
 
 class EppTestCase < ActionDispatch::IntegrationTest
   include Assertions::EppAssertions
+
+  def return_xml_domain_schema_version(response_xml)
```
yulgolem, line 72: I'd suggest to modify this method with adding filename prefix, ATM it's just hardcoded as a `domain`, so we could not use this method to check for, say, contact schema version.

test/integration/epp/domain/info/base_test.rb
```diff
@@ -25,7 +25,11 @@ def test_returns_valid_response
          headers: { 'HTTP_COOKIE' => 'session=api_bestnames' }
 
     response_xml = Nokogiri::XML(response.body)
+
+    schema_version = return_xml_domain_schema_version(response_xml)
+   
     assert_epp_response :completed_successfully
+    assert schema_version >= 1.1
```
yulgolem, line 32: I'd suggest to create dedicated assertion method using return_xml_domain_schema_version(response_xml) as a private method call, so instead of two lines 
```
schema_version = return_xml_domain_schema_version(response_xml)
assert schema_version >= 1.1
```
we could just do something like
```
assert schema_version_bigger_for('domain', 1.1)
```

yulgolem (changes requested):

yulgolem (approved):

yulgolem (changes requested):
That's it!
```
Anytime the version of any imported schema is raised, the version of 'all' schema is also raised.
```

So instead of changing link to the domain-eis-1.1.xsd in all-ee-1.1.xsd, we should:
1) Introduce new all-ee-1.2.xsd schema
1) Also make change in `config/initializers/load_schemas.rb`

yulgolem (approved):

vohmar:
one request - please rename the domain-eis-1.1.xsd to domain-ee-1.1.xsd (including targetNamespaces in the file), that would also mean changing the link in all-ee-1.2.xsd. I also noticed that in all-ee-1.2.xsd the schema targetNamespace is not updated

vohmar:
apparently the new schema versions are not applied automatically.

header of domain info response:
  <response>
    <result code="1000">
      <msg>Command completed successfully</msg>
    </result>
    <resData>
      <domain:infData xmlns:domain="https://epp.tld.ee/schema/domain-eis-1.0.xsd">

same goes for sample queries in xml console of registrar's portal and portal ui requests. If I tried to replace the schema versions manually in the request I was returned with ugly html error (we fixed that one once didn't we?)

sample request:
  <command>
    <info>
      <domain:info
       xmlns:domain="https://epp.tld.ee/schema/domain-ee-1.1.xsd">
        <domain:name hosts="all">10milli.ee</domain:name>
        <domain:authInfo>
          <domain:pw>2fooBAR</domain:pw>
        </domain:authInfo>
      </domain:info>
    </info>
    <clTRID>Märi Änn R1-1622818050</clTRID>
  </command>
</epp>

app/controllers/registrar/xml_consoles_controller.rb
```diff
@@ -2,13 +2,14 @@ class Registrar
...
-    def show
-    end
+    def show; end
 
     def create
       begin
         @result = depp_current_user.server.request(params[:payload])
-      rescue
+
+        checking_schema_valid_path(params[:payload])
```
yulgolem, line 11: I'd suggest to rename it to `check_schema_path`

app/controllers/registrar/xml_consoles_controller.rb
```diff
@@ -18,8 +19,82 @@ def load_xml
       cl_trid = "#{depp_current_user.tag}-#{Time.zone.now.to_i}"
       xml_dir_path = Rails.root + 'app/views/registrar/xml_consoles/epp_requests'
       xml = File.read("#{xml_dir_path}/#{params[:obj]}/#{params[:epp_action]}.xml")
-      xml.gsub!('<clTRID>ABC-12345</clTRID>', "<clTRID>#{cl_trid}</clTRID>")
+      xml = insert_schema_path(xml, cl_trid)
```
yulgolem, line 22: We are not only inserting schema path here, but cl_trid too, so I'd suggest to rename method to something like `prepare_payload`

app/controllers/registrar/xml_consoles_controller.rb
```diff
@@ -18,8 +19,82 @@ def load_xml
...
+      path = regex_to_find_poll_schema(payload) if regex_to_find_poll_schema(payload).present?
+
+      @result = template_wrong_path unless array_valid_paths.include? path
+    end
+
+    def array_valid_paths
+      Xsd::Schema::PREFIXES.map { |prefix| Xsd::Schema.filename(for_prefix: prefix) }
+    end
+
+    def template_wrong_path
```
yulgolem, line 42: Well, why not a `wrong_path_response`?

app/controllers/registrar/xml_consoles_controller.rb
```diff
@@ -18,8 +19,82 @@ def load_xml
...
+
+    def regex_to_find_poll_schema(payload)
+      contact_schema_tag = payload.scan(/poll\S+/)
+      return if contact_schema_tag.empty?
+
+      'https://epp.tld.ee/schema/epp-ee-1.0.xsd'
+    end
+
+    def insert_schema_path(xml, cl_trid)
+      xml.gsub!('xmlns:domain="domain-ee"',
```
yulgolem, line 87: How about doing an array of prefixes and iterate through this array with gsub for a current prefix?

yulgolem (changes requested):

yulgolem (approved):

PR #2011, opened 2021-05-28
Mark internal invoices as paid prior to sending

### LINKED ISSUE #2009: unpaiable invoices are not marked as paid and should also be sent as e-invoice
If registry is unable to match received payment to any existing unpaid invoice for that particular registrar (indicated by reference nr) new invoice is generated and instantly marked as paid. The invoice is generated with unpayable label and sent to the registrar only via email.

* the unpayable invoice must have receipt date filled. As an additional bonus we might add Paid status on the PDF invoice
* the unpayable invoices should also be sent as e-invoice but these invoices must also have the Payable tag set to NO

yulgolem:
Closes #2009

### DIFF
--- a/app/jobs/send_e_invoice_job.rb
+++ b/app/jobs/send_e_invoice_job.rb
@@ -21,7 +21,7 @@ class SendEInvoiceJob < ApplicationJob
   end
 
   def process(invoice:, payable:)
-    invoice.to_e_invoice(payable: payable).deliver
+    invoice.to_e_invoice(payable: payable).deliver unless Rails.env.development?
     invoice.update(e_invoice_sent_at: Time.zone.now)
     log_success(invoice)
   end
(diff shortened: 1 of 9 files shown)


yulgolem:
OK, looks like it works.

To manually test on staging we should run
```
RAILS_ENV=staging bundle exec rake invoices:test_payments
```

The task will create a dummy payment dataset for a buyer of last invoice. Since we do not have unpaid invoice atm, the task will create invoice/e-invoice as well and print their data.

Most interesting data here is
```
e_invoice_sent_at: Mon, 31 May 2021 13:38:16.889503000 EEST +03:00>
Last invoice paid at 2021-05-31
```

date of e-invoice sent and invoice receipt date. Already paid invoice/pdf could also be checked at Mailcatcher service.

PR #2019, opened 2021-06-02
Automate checking epp responses against  schema

### LINKED ISSUE #1930: Automate checking response against schema
Using approach presented at https://github.com/internetee/registry/pull/1929 we should automate testing EPP response against schema in order to keep XML schemas up-tp-date.

Maybe we should add response checks to some helpers like `assert_epp_response :completed_successfully`.

yulgolem:
Closes #1930

### DIFF
--- a/app/views/epp/domains/info.xml.builder
+++ b/app/views/epp/domains/info.xml.builder
@@ -41,7 +41,7 @@ xml.epp_head do
 
         xml.tag!('domain:clID', @domain.registrar.code)
 
-        xml.tag!('domain:crID', @domain.cr_id)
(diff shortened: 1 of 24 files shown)


app/views/epp/domains/info.xml.builder
```diff
@@ -41,7 +41,7 @@ xml.epp_head do
 
         xml.tag!('domain:clID', @domain.registrar.code)
 
-        xml.tag!('domain:crID', @domain.cr_id)
+        xml.tag!('domain:crID', @domain.cr_id) if @domain.cr_id.present?
```
yulgolem, line 44: otherwise we might show it if it's nil and that's schema type violation

PR #2028, opened 2021-06-10
Add request throttling

### LINKED ISSUE #1017: Unified rate limiting system
We need to come up with a solution to make unified query rate limiting system.
Current rate limiting system forces the application to one machine and is far from ideal for ipv6.

Every registrar has a set limit x of allowed queries.
Every request to EPP or REPP should go to one token bucket.
Rate limiting should be configurable per registrar if needed ( disable or amount ) 
Rate limiting system should emit metrics about query load globally and per registrar, 
failed and completed queries.

The rate, query and session limiting should be done on application level instead of FW level like its done now.

### LINKED ISSUE #2270: Add ratelimiting to all epp and repp operations
Related to: https://github.com/internetee/registry/issues/1017
Use this: https://github.com/internetee/registry/pull/2028

New rate limiting solution is currently not applied to EPP poll, login and logout + any REPP requests. Rate limiting must be extended to all EPP and REPP requests. It might be good idea to exclude logout though so registrars are always able to end their sessions. Instead error rate limiting could be used here instead to stop spamming of logout requests i.e. too many erroneous requests per minute or second

yulgolem:
Closes #1017
close #2270


yulgolem:
How to test:

There are environment variables to be set. All variable values are strings.
`ENV["shunter_default_timespan"]` is a check timespan in seconds, default "60".
`ENV["shunter_default_threshold"]` is a number of requests per timespan allowed, default "100".
`ENV["shunter_enabled"]` is gem enabled or not, possible string values `true` or `false`, if not set, considered `false`.

Actions to be limited are domain/contact `info, renew, update, transfer, delete`

So - set minimum `ENV["shunter_default_threshold"]` (e.g. `"2"`) and try to get three domain info requests in 60 seconds.

ratM1n:
redis server location is hardcoded? it will search for "redis:6379", can this be changed from configuration?
returned error xml doesn't contain the error message, only the error number. i think message should contain information about the current limits eg.
`rate limit exceeded, current limit is 10 queries in 60 sec.`

queries ending with error (eg. query against the non existing objects) are not counted and limited

yulgolem:
Unfortunately due to architectural limitations we are unable to rate limit erroneous requests - they are (like unauthorised/not found) processed before we are know anything about user to limit (unauthorised error) or on routing level before even entering application (not found).

Redis connection is also configurable.
ENV["shunter_redis_host"] and ENV["shunter_redis_port"]
ENV["shunter_default_timespan"] is a check timespan in seconds, default "60".
ENV["shunter_default_threshold"] is a number of requests per timespan allowed, default "100".
ENV["shunter_enabled"] is feature enabled or not, possible string values true or false, if not set, considered false.

Actions to be limited are domain/contact info, renew, update, transfer, delete

So - set minimum ENV["shunter_default_threshold"] (e.g. "2") and try to get three domain info requests in 60 seconds.
The response should be of 2502 session_limit_exceeded_server_closing_connection with msg describing current limits.

DinSmol (approved):

vohmar:
It looks like the limit does not apply to login, logout and poll requests and REPP requests. While logout seems really a good choice to exclude from the rate limiting then login and poll should definitely be included and same goes for REPP - rate limit should apply per registrar and overlapping both epp and repp services

yulgolem:
> It looks like the limit does not apply to login, logout and poll requests and REPP requests.
yep, it's only added to info, renew, update, transfer, delete actions of domain and contacts epp controller as a proof on concept.

if the concept will be approved any other controller actions could be throttled as well, but I suppose that's out of scope for this PR.

thiagoyoussef:
Spreadsheet showing epp and repp actions that uses rate limiting. Also added which ones are being tested.

[Rate Limiting.xlsx](https://github.com/internetee/registry/files/9277069/Rate.Limiting.xlsx)

thiagoyoussef:
I added tests for several actions. Basically is the same test, but requesting for a different action. So maybe It would be better to remove a few tests to avoid repeating.
What do you think is best?

vohmar:
Something is off with this - maybe misconfigured, maybe something else. Only thing request that works is login all the others return the same error:

```
got input: 204 bytes.
  <command>
    <poll op="req"></poll>
    <clTRID>QueryNO-1</clTRID>
  </command>
</epp>

EPP/TCP Header: expecting 388 bytes of XML
  <response>
    <result code="2400">
      <msg lang="en"/>
    </result>
    <trID>
      <clTRID>QueryNO-1</clTRID>
      <svTRID>ccReg-2647485945</svTRID>
    </trID>
  </response>
</epp>
```

Server conf:

```
shunter_enabled: "true"
shunter_default_timespan: "60"
shunter_default_threshold: "10"
```

PR #2057, opened 2021-06-22
request and response have the same domain schema

### LINKED ISSUE #2046: EPP: responses must reference the same schema version that was used in valid request
related to https://github.com/internetee/registry/issues/2030

with support for multiple schema file versions we need also make sure we are responding with the same schema file paths that were used in valid request

OlegPhenomenon:
Close #2046

### DIFF
--- a/app/controllers/epp/domains_controller.rb
+++ b/app/controllers/epp/domains_controller.rb
@@ -4,6 +4,7 @@ module Epp
     before_action :find_domain, only: %i[info renew update transfer delete]
     before_action :find_password, only: %i[info update transfer delete]
     before_action :set_paper_trail_whodunnit
+    before_action :parse_schemas_prefix_and_version
 
     def info
       authorize! :info, @domain
@@ -242,6 +243,15 @@ module Epp
         return false
       end
       true
+    end
+
+    def parse_schemas_prefix_and_version
+      return unless params[:frame]
+
+      xml = params[:frame].gsub!(/(?<=>)(.*?)(?=<)/, &:strip).to_s
+      res = xml.match(/xmlns:domain=\"https:\/\/epp.tld.ee\/schema\/(?<prefix>\w+-\w+)-(?<version>\w.\w).xsd/)
+      @schema_version = res[:version]
+      @schema_prefix = res[:prefix]
     end
   end
 end
--- a/app/views/epp/domains/create.xml.builder
+++ b/app/views/epp/domains/create.xml.builder
@@ -5,7 +5,8 @@ xml.epp_head do
     end
 
     xml.resData do
-      xml.tag!('domain:creData', 'xmlns:domain' => Xsd::Schema.filename(for_prefix: 'domain-ee')) do
+      xml.tag!('domain:creData',
+               'xmlns:domain' => Xsd::Schema.filename(for_prefix: @schema_prefix)) do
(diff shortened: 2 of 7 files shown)


app/controllers/epp/domains_controller.rb
```diff
@@ -4,6 +4,7 @@ class DomainsController < BaseController
     before_action :find_domain, only: %i[info renew update transfer delete]
     before_action :find_password, only: %i[info update transfer delete]
     before_action :set_paper_trail_whodunnit
+    before_action :parsing_schemas_prefix_and_version
```
yulgolem, line 7: Method name should be without `-ing` suffix, so just `parse_schema_prefix_and_version` will do.

app/controllers/epp/domains_controller.rb
```diff
@@ -22,7 +23,7 @@ def info
       when 'all'
         @nameservers = @domain.nameservers.sort
       end
-
+
```
yulgolem, line 26: Please, remove these whitespaces.

app/views/epp/domains/create.xml.builder
```diff
@@ -5,7 +5,8 @@ xml.epp_head do
     end
 
     xml.resData do
-      xml.tag!('domain:creData', 'xmlns:domain' => Xsd::Schema.filename(for_prefix: 'domain-ee')) do
+      xml.tag!('domain:creData',
+               'xmlns:domain' => Xsd::Schema.filename(for_prefix: @schema_prefix)) do
```
yulgolem, line 9: And what about version? As far as I remember `Xsd::Schema.filename(for_prefix: <some_prefix>))` still returns latest version of a filename.

yulgolem (changes requested):

yulgolem:
I suppose we still need to create a test to check if we are returning not the latest schema version, but the version from the request if that's supported. ATM I can't find any.

OlegPhenomenon:
> I suppose we still need to create a test to check if we are returning not the latest schema version, but the version from the request if that's supported. ATM I can't find any.

test/integration/epp/doman/info/base_test.rb

test_returns_valid_response_if_schema_version_is_previous

yulgolem:
> > I suppose we still need to create a test to check if we are returning not the latest schema version, but the version from the request if that's supported. ATM I can't find any.
> 
> test/integration/epp/doman/info/base_test.rb
> 
> test_returns_valid_response_if_schema_version_is_previous

That's it! We are getting not the response we need, but a last one - since we only got domain-eis-1.0.xsd .
Try to add  domain-eis-1.1.xsd and check if the response persists.

app/views/epp/domains/create.xml.builder
OlegPhenomenon: Related #2058

OlegPhenomenon:
> > > I suppose we still need to create a test to check if we are returning not the latest schema version, but the version from the request if that's supported. ATM I can't find any.
> > 
> > 
> > test/integration/epp/doman/info/base_test.rb
> > test_returns_valid_response_if_schema_version_is_previous
> 
> That's it! We are getting not the response we need, but a last one - since we only got domain-eis-1.0.xsd .
> Try to add domain-eis-1.1.xsd and check if the response persists.

Related #2058

PR #2061, opened 2021-06-25
make supporting version for schema

### LINKED ISSUE #2058: Make schema versioning
In this case, when we call some specific schema with a specific prefix, the registry takes the most recent version of the file. If in the future we update the schemas, for example, from 1.1 to 1.2, and we have to implement support for old versions again, then this will not be easy, since the most recent version will be called again.

It is necessary to make more flexible the ability to call schemas, where it will be called not only by the schema, but also by the version.

OlegPhenomenon:
Close #2058

### DIFF
--- a/lib/xsd/schema.rb
+++ b/lib/xsd/schema.rb
@@ -3,8 +3,8 @@ module Xsd
     SCHEMA_PATH = 'lib/schemas/'.freeze
     BASE_URL = 'https://epp.tld.ee/schema/'.freeze
 
-    REGEX_PREFIX_WITH_DASH = /(?<prefix>\w+-\w+)-(?<version>\w.\w).xsd/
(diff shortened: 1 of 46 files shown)


lib/xsd/schema.rb
```diff
@@ -19,11 +22,12 @@ class Schema < ApplicationService
       secDNS
     ].freeze
 
-    attr_reader :xsd_schemas, :for_prefix
+    attr_reader :xsd_schemas, :for_prefix, :for_version
 
     def initialize(params)
       schema_path = params.fetch(:schema_path, SCHEMA_PATH)
       @for_prefix = params.fetch(:for_prefix)
+      @for_version = params[:for_version] || '1.1'
```
yulgolem, line 30: How about `@for_version = params.fetch(:for_version, '1.1')`? https://apidock.com/rails/v4.0.2/ActionController/Parameters/fetch

It will help to maintain consistent codestyle, I suppose.

yulgolem (approved):

PR #2063, opened 2021-06-28
WIP Interim save

yulgolem:


yulgolem:
- let's create an event storage, store there 1) email validation results and 2) admin set FDs
- then let's initialize FD from these events

And for the next PRs:
- then let's add business registry validations, since we got no such a functionality in a production yet, just a manually working prototype
- then we'll setup not only automatic FD from events, but FD lifting as well when all negative events are gone
- and just after that - we'll think about thresholds

yulgolem:
Rake task is started as `bundle exec rake verify_email:check_all -- -d=shop.test --check_level=mx --spam_protect=true`
or `bundle exec rake verify_email:check_all -- -dshop.test -cmx -strue`

Issue #1795, opened 2021-01-11
Registrar: Validation contacts name - Special characters instead of letters


OlegPhenomenon:
If during the creation of a new contact you type a different number of characters, for example ¤ # "¤ #" ¤ "?,then the form allows you to create a new contact with that name.

yulgolem:
BTW FYI https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

vohmar:
fixed by https://github.com/internetee/registry/pull/2074

yulgolem:
Reopened (PR reverted) since one of our clients with name containing "&"-character could not make any changes to the contact name.
Asked Silver on legal limitations we have on this account, let's set this task on hold until we got a knowledge on what characters should we allow/block.

yulgolem:
According to [technical terms and conditions ](https://meedia.internet.ee/files/Appendix%201_Registrar_Contract_EN_01.07.2020.pdf)of our contracts (courtesy of Silver Sõrmus) we got no legal restrictions on any naming whatsoever - so even name of `¤ # "¤ #" ¤ "?` is completely legal. So my proposal is to close this issue.

PR #2074, opened 2021-07-07
fix contacts name validation

DinSmol:
see issue #1795


yulgolem (changes requested):
BTW that's the real contact name - `MARY ÄNN O’CONNEŽ-ŠUSLIK`. And it doesn't matches with this regexp. Eesti keelt uses all types of diacritic symbols/dashes in person names.

yulgolem (approved):

Issue #1788, opened 2021-01-06
Admin - Add new bank transaction: undefined method `paid?' for nil:NilClass 


OlegPhenomenon:
When trying to bind manually new bank transaction in admin panel, there is an error: Add new bank transaction: undefined method `paid?' for nil:NilClass 

Reproducing the error step by step:
1. Go settings and click to Invoice
2. Select some unpaid invoice
3. Click to "Payment received"
4. Click to save
5. Click to add in Bank Transaction sections
6. Fill in the fields
7. Left empty field binded invoice or type something and click "Bind manually"

vohmar:
fixed by #2075

PR #2075, opened 2021-07-07
fix bind_invoice method

DinSmol:
see issue #1788

### DIFF
--- a/app/models/bank_transaction.rb
+++ b/app/models/bank_transaction.rb
@@ -106,7 +106,7 @@ class BankTransaction < ApplicationRecord
   private
 
   def reset_pending_registrar_balance_reload
-    return unless registrar.present? && registrar.settings['balance_auto_reload']
+    return unless registrar.settings['balance_auto_reload']
 
     registrar.settings['balance_auto_reload'].delete('pending')
     registrar.save!
--- a/app/models/bank_transaction.rb
+++ b/app/models/bank_transaction.rb
@@ -106,7 +106,7 @@ class BankTransaction < ApplicationRecord
   private
 
   def reset_pending_registrar_balance_reload
-    return unless registrar.settings['balance_auto_reload']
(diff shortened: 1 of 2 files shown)


app/models/bank_transaction.rb
```diff
@@ -60,7 +60,7 @@ def bind_invoice(invoice_no, manual: false)
 
     invoice = Invoice.find_by(number: invoice_no)
     errors.add(:base, I18n.t('invoice_was_not_found')) unless invoice
```
yulgolem, line 62: Let's consider to move this check also to` #validate_invoice_data(invoice)` method.

yulgolem (approved):

yulgolem (changes requested):
Let's add some tests to be sure.

yulgolem (approved):

vohmar:
might be different issue, but maybe its and easy fix to do right here. the method paid error is indeed gone, but if a user creates a valid bank transaction that matches an unpaid invoice error is returned on pressing "bind manually"

undefined method `settings' for nil:NilClass

app/models/bank_transaction.rb
```diff
@@ -102,7 +106,7 @@ def parsed_ref_number
   private
 
   def reset_pending_registrar_balance_reload
-    return unless registrar.settings['balance_auto_reload']
+    return unless registrar.present? && registrar.settings['balance_auto_reload']
```
yulgolem, line 109: Looks like we do not have this covered in tests, are we?

yulgolem (changes requested):

yulgolem:
I'd suggest to move a bind manually error `undefined method `settings' for nil:NilClass` to separate issue and complete/close this one - you know, just to eat the elephant one bite a time.

yulgolem (approved):

PR #2081, opened 2021-07-08
Fix for contact birth date validation

### LINKED ISSUE #1796: Registrar: Validation birthday field in Contact - Allow a non-existent date of birth
Steps:
1. Go to contacts and create a contact
2. Choose any country instead of Estonia
3. In the type field select the date of birth
4. Type tomorrow or any future day in the birthday field

The form creates a contact with a date of birth that will happen in the future

UPDATE:
Also the form skips non-existent dates, for example 0000-34-35

DinSmol:
close #1796

### DIFF
--- a/app/models/epp/contact.rb
+++ b/app/models/epp/contact.rb
@@ -8,6 +8,9 @@ class Epp::Contact < Contact
   # disable STI, there is type column present
   self.inheritance_column = :sti_disabled
 
+  VALID_BIRTH_DATE_FROM = Time.zone.today - 150.years
+  VALID_BIRTH_DATE_TO = Time.zone.tomorrow
+
   before_validation :manage_permissions
 
   def manage_permissions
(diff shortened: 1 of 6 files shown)


app/models/epp/contact.rb
```diff
@@ -16,6 +16,26 @@ def manage_permissions
...
+    return unless Depp::Contact::SELECTION_TYPES[2].include?(ident_type)
+
+    if (Date.parse(ident) rescue ArgumentError) == ArgumentError
+      add_epp_error('2308', nil, nil, I18n.t('epp.contacts.errors.valid_ident_date_format'))
+      @error = true
+      return
+    end
+
+    contact_ident_date = Date.parse(ident)
+    valid_time_range = (Date.today - 125.years)...(Date.tomorrow - 18.years)
```
yulgolem, line 31: Let's save it in some constant or Value Object.
And why 18 years ago? Do we have any legal regulations on that?

yulgolem (changes requested):

app/models/epp/contact.rb
DinSmol: I am not aware of any legal regulations regarding this. This is just my guess. I'll ask Timo about it.

yulgolem (approved):

PR #2083, opened 2021-07-08
added migration and new json field for store domain statuses history

### LINKED ISSUE #2080: Implement new field for store statuses in domain model
Related #https://github.com/internetee/registry/pull/1902
Related https://github.com/internetee/registrant_center/issues/93

In #1902 said the following:
```
PR was reverted due to high amount of DB load on data migrations.
Set on hold until Sidekiq will be online in prod so we could implement the PR in question in two parts:

    Preparing the DB via async jobs
    Updating code itself.

```

Therefore, all the fields that were created in the #1902 case disappeared and the ticket with the preservation of domains after unlocking becomes relevant.
Since we have had changes in the structures of the project, the plan is as follows:
1. This PR will have to go to the master FIRST! because it prepares the structure for data migration.
2. The second step will be to create a rake task (this will be a separate PR), which will migrate data to a new database structure
3. But the third step and the third PR will decide the question of how to maintain the necessary statuses

OlegPhenomenon:
Close #2080


app/jobs/migrate_before_force_delete_statuses_job.rb
```diff
@@ -0,0 +1,9 @@
+class MigrateBeforeForceDeleteStatusesJob < ApplicationJob
+  def perform
+    domains = Domain.where.not(statuses_before_force_delete: nil)
+    domains.each do |domain|
```
yulgolem, line 4: I'd suggest use find_each here or smth working in batches.

app/jobs/migrate_statuses_to_domain_history_job.rb
```diff
@@ -0,0 +1,9 @@
+class MigrateStatusesToDomainHistoryJob < ApplicationJob
+  def perform
+    domains = Domain.all.select { |d| !d.locked_by_registrant? }
```
yulgolem, line 3: How about create some scope?
a.select is not a cheap operation.

yulgolem (changes requested):

app/jobs/migrate_before_force_delete_statuses_job.rb
```diff
@@ -0,0 +1,9 @@
...
+      domain.force_delete_domain_statuses_history = domain.statuses_before_force_delete
+      domain.save
```
yulgolem, line 6: How about look for domains not only `.where.not(statuses_before_force_delete: nil)`, but `where(force_delete_domain_statuses_history: nil)`? That could help us keep task idemponent.

OlegPhenomenon: yes, good idea

yulgolem:
Got an error trying to install on staging:
```
PG::DuplicateColumn: ERROR:  column \"json_statuses_history\" of relation \"domains\" already exists"
```

Suppose that's because first PR regarding this feature was deployed but migration wasn't rolled back.
Could you, please, try to deploy this PR on staging?

yulgolem:
1) Logging on staging does not work.
1) Every domain save calls for a UpdateWhoisRecordJob started. Consider not saving, but updating attribute on a domain, thus bypassing the callbacks.

yulgolem (approved):

OlegPhenomenon:
Deploy notes:
**After merging,** it needs to run the task `migrate_domain_statuses:fd_domains` to copy statuses before forced deletion, as well as `migrate_domain_statuses: admin_status_history`

It also need to run these tasks **after release and deployment.** It need to wait until the task completes its work and there are no scheduled jobs left in the queue. The task will have to return zero processed domains.

yulgolem:
BTW @OlegPhenomenon , say we'll run tasks, copy statuses before FD... then some domains will be force deleted.
Don't you think we'll lose statuses from these domains which will be FDed _after_ the jobs was run?

OlegPhenomenon:
> BTW @OlegPhenomenon , say we'll run tasks, copy statuses before FD... then some domains will be force deleted.
> Don't you think we'll lose statuses from these domains which will be FDed _after_ the jobs was run?

If they are deleted, then I see no reason to worry :) because if they are deleted, then in general all the data of this domain will be deleted :) The data is not saved in a separate model, but in a new field of each domain.

yulgolem:
> > BTW @OlegPhenomenon , say we'll run tasks, copy statuses before FD... then some domains will be force deleted.
> > Don't you think we'll lose statuses from these domains which will be FDed _after_ the jobs was run?
> 
> If they are deleted, then I see no reason to worry :) because if they are deleted, then in general all the data of this domain will be deleted :) The data is not saved in a separate model, but in a new field of each domain.

nope, they will not be deleted - we'll just lost their statuses_before_fd data, since it will be stored in the old field, but from new PR you will be looking for a new one.

PR #2097, opened 2021-07-15
Fix for visibility domain delete buttons (after domain deleting)

### LINKED ISSUE #1798: Registrar: Domain - No any message that domain was deleted
When I deleted some domain, it returns me to the back page and don't show me any message, also delete button it still available

DinSmol:
close #1798


yulgolem:
Since we need to test on registrar portal integration side, I'd suggest to pass it - it's too difficult atm to stub the epp requests to check UI.
Manual testing was OK.

yulgolem (approved):

PR #2126, opened 2021-08-19
Admin: the functionality to change registrar credit balance

### LINKED ISSUE #714: Admin: interface for administrative credit returns and corrections
Currently the only way to fix credit issues is directly changing data in the db - error prone. Instead there should be an interface to make such changes with option to leave comment and remove the risk of human error

goal: add an option to administratively change registrar's credit balance
https://st-adm.infra.tld.ee/admin/registrars/1

while changing the balance there needs to be an option to add a comment/explanation. Update has to create an account activity record with the comment in description field and activity type update credit

https://st-adm.infra.tld.ee/admin/account_activities

DinSmol:
close #714


test/system/admin_area/accounts_test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    assert_response :ok
+    assert_equal 'text/csv; charset=utf-8', response.headers['Content-Type']
+    assert_equal %(attachment; filename="accounts_#{Time.zone.now.to_formatted_s(:number)}.csv"; filename*=UTF-8''accounts_#{Time.zone.now.to_formatted_s(:number)}.csv),
+                 response.headers['Content-Disposition']
+    assert_not_empty response.body
+  end
+
+  def test_change_account_balance
+    puts @account.inspect
```
OlegPhenomenon, line 23: Remove this line. It doesn't make sense

app/controllers/admin/accounts_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+                                     currency: @account.currency,
+                                     description: params[:description],
+                                     activity_type: AccountActivity::UPDATE_CREDIT)
+
+      if activity.save
+        true
+      else
+        false
+      end
+    end
```
OlegPhenomenon, line 40: Small mark, although not essential. Just for the beauty of the code
```
return true if activity.save
false
```

OlegPhenomenon (changes requested):

app/controllers/admin/accounts_controller.rb
OlegPhenomenon: Although I could even call like `save!`, so that if it fails to save, it throws an exception about the reasons for the error. Then there is no need for conditions. It seems to me that if it returns false, it will take time trying to debug the code.

yulgolem: Well, that could be done like that, but not there.

Considered a good habit to move all stuff not related to action rendering (i.e. business logic) from controllers to the service layer- to service objects/interactors etc. See David Copeland, Sustainable Web Development with Ruby on Rails (https://www.amazon.com/Sustainable-Web-Development-Ruby-Rails-ebook/dp/B08ND6X37Q)

yulgolem (changes requested):

app/controllers/admin/accounts_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Admin
+  class AccountsController < BaseController
+    load_and_authorize_resource
+
+    def index
+      @q = Account.includes(:registrar).search(params[:q])
```
yulgolem, line 6: We are creating strong params on `account_params`, but using not protected `params` all the way?

yulgolem (approved):

OlegPhenomenon (approved):

PR #2150, opened 2021-09-07
Ruby3 support

DinSmol:
Fixes for ruby 3 supporting

### DIFF
--- a/app/models/epp/domain.rb
+++ b/app/models/epp/domain.rb
@@ -173,7 +173,7 @@ class Epp::Domain < Domain
 
   ### RENEW ###
 
-  def renew(renewed_expire_time:, period:, unit:)
+  def renew(renewed_expire_time, period, unit)
     @is_renewal = true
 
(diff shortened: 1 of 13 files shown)


yulgolem:
@DinSmol could you, please, also add to /.github/workflows/ruby an auto-test against ruby 3.0?

app/models/epp/domain.rb
```diff
@@ -173,7 +173,7 @@ def set_pending_delete!
 
   ### RENEW ###
 
-  def renew(renewed_expire_time:, period:, unit:)
+  def renew(renewed_expire_time, period, unit)
```
yulgolem, line 176: Move from keyword arguments to simple ones?

yulgolem (changes requested):
@DinSmol Also take a look, please, on /lib/monkey_patches/i18n.rb

yulgolem (approved):

PR #2153, opened 2021-09-07
added dnskey update prohibited status

### LINKED ISSUE #2143: Add an exception for DNSSEC in the registry lock
Current implementation of registry lock denies all changes to the domain registration data except renewals, expiration and administrative actions like forceDelete

prohibiting any kind of updates to the domain registration raises issues with DNSSEC though where in case of hosted solutions the service provider is expected to roll the keys regularly. To keep DNSSEC still possible in the registry locked state we should add an option to enable dnssec key rollovers while locked

this would mean adding new special status to use instead of *UpdateProhibited status like with serverRegistrantChangeProhibited. Perhaps create special registry lock status or statuses.

by default locking domain should leave extension:secDNS element out of the restriction scope with an option for domain owner to enable limitations to that element as well. 

by default setting registry lock on a domain restrict following epp requests:
* domain:update operatsion with exception of extension:secDNS element
* domain:transfer
* domain:delete

allowed operatsions by default:
* domain:renew
* domain:update for only extension:secDNS element

If user chooses to lock dnssec as well then locking domain restricts:
* domain:update
* domain:transfer
* domain:delete

and allows 
* domain:renew

OlegPhenomenon:
Close #2143


yulgolem (approved):

vohmar:
objUpdateProhibited
* epp error message: <result code="2400"><msg lang="en"/>
* adding new nameserver and contact objects to the domain succeeded: https://st-adm.infra.tld.ee/admin/epp_logs/46939
* removing existing nameserver and contact objects was successful: https://st-adm.infra.tld.ee/admin/epp_logs/46963
* registrant change was successful: https://st-adm.infra.tld.ee/admin/epp_logs/47027

so the only operation that failed was changing ip addresses of existing nameserver records. Adding new dnssec key was successful as expected

extensionUpdateProhibited

seems to work as expected
changing ns ip successful, dnssec key removal, adding and change was denied

vohmar:
serverObjUpdateProhibited - spotted issues:

* on adding ipv6 address to an existing ns record this error was returned:  <result code="2304"><msg lang="en">Object status prohibits operation :serverDnskeyUpdateEnabled</msg> - there is no such status, the message should reference serverObjUpdateProhibited in this scenario (https://st-adm.infra.tld.ee/admin/epp_logs/47423). 
  * same error message on adding new ns record, removing ns record, registrant change
  
* in admin where new statuses can be set on an object the new statuses are listed in a complete form while other statuses have server prefix omitted  as one can only set server prefixed statuses in admin. Please do the same for the new statuses

yulgolem (approved):

app/models/feature.rb
```diff
@@ -0,0 +1,6 @@
+class Feature
+  def self.obj_and_extensions_statuses_enabled?
+    return false if ENV['obj_and_extensions_prohibited'] == 'false'
```
yulgolem, line 3: Empty line here

app/models/feature.rb
```diff
@@ -0,0 +1,6 @@
...
+    ENV['obj_and_extensions_prohibited'] || false
+  end
+end
```
yulgolem, line 6: And here

yulgolem (changes requested):
Let's add some high-level tests for the feature-toggle.
Like 'key is absent or set to false - feature not active, key. is present and set to true - feature is active'

yulgolem (approved):

PR #2161, opened 2021-09-16
REPP: bulk nameservers addition

### LINKED ISSUE #2158: REPP: bulk ns change - add an option to add new ns
registrar has requested for an option to add new nameserver record to many domains. 

Maybe we could solve this by making the id parameter optional in the bulk ns change request
https://internetee.github.io/repp-apidoc/#bulk-nameserver-change

so if id is missing or blank than nameservers from the attributes block is just added to the list of domains

Requires changes in REPP and repp-doc + possibly in registrar

DinSmol:
close #2158

PUT /repp/v1/registrar/nameservers
{
    "data": {
        "type": "nameserver",
        "domains": ["only.ee", "these.ee", "domains.ee"],
        "attributes": {
            "hostname": "ns2.ams1.domainer.ee",
            "ipv4": ["192.168.1.1"],
            "ipv6": ["2620:119:35::36"]
        }
    }
}

As result nameserver "ns2.ams1.domainer.ee" will be add to next domains: "only.ee", "these.ee", "domains.ee"

### DIFF
--- a/app/models/registrar.rb
+++ b/app/models/registrar.rb
@@ -43,6 +43,7 @@ class Registrar < ApplicationRecord
   after_commit :update_whois_records
   def update_whois_records
     return true unless changed? && (changes.keys & WHOIS_TRIGGERS).present?
+
     RegenerateRegistrarWhoisesJob.perform_later id
   end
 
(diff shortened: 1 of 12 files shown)


app/models/registrar.rb
```diff
@@ -175,6 +176,36 @@ def replace_nameservers(hostname, new_attributes, domains: [])
...
+  def add_nameservers(new_attributes, domains: [])
+    transaction do
+      domain_scope = domains.dup
+      domain_list = []
+      failed_list = []
+
+      return if domains.empty?
+
+      domain_scope.each do |domain_name|
+        domain = self.domains.find_by('name = ? OR name_puny = ?', domain_name, domain_name)
```
yulgolem, line 188: How about putting that block in the separate method to lower the complexity/lines?

app/models/registrar.rb
```diff
@@ -175,6 +176,36 @@ def replace_nameservers(hostname, new_attributes, domains: [])
...
+
+        if domain.blank? || domain_not_updatable?(hostname: new_attributes[:hostname], domain: domain)
+          failed_list << domain_name
+          next
+        end
+
+        new_nameserver = Nameserver.new
```
yulgolem, line 195: And that block as well.

yulgolem (changes requested):

yulgolem (approved):

PR #2175, opened 2021-10-04
Fix for DNSSEC keys updating

### LINKED ISSUE #2174: Update of DNSSEC keys fails with registry lock set
Registry lock set by default only serverObjUpdateProhibited status affecting domain update requests. This status should not affect extensions to update request, but dnssec still fails.

Further investigation revealed that update requests generated by registrar portal were successful but requests directly to API tend to fail.

DinSmol:
close #2174


yulgolem:
@DinSmol could you, please, also add a test simulating situation in question?

PR #2199, opened 2021-10-25
Update company name if it doesn't match with name in registry

OlegPhenomenon:


vohmar:
please change the poll message

Contact update: contact_id name updated from prev_name to new_name by the registry

PR #2201, opened 2021-10-27
Optimize verify email

OlegPhenomenon:
**Problem:** every time contacts are verified, verification success records are written to the database, which increases the size of the database. Given the number of contacts, the database will grow significantly.

**Solution:**
So far I have done it in a simple way:
There is a certain time after which contacts that have been verified are considered outdated. This period is set by the VALIDATION_PERIOD constant. **For testing purposes, I have set one day for now, then when we make sure that everything works as it should, we will need to change the value of this constant.**

After that, I take contacts that, firstly, have not been verified yet, and also take contacts that have passed verification, but by the date of verification they are considered outdated.

And now these contacts are re-verified.

Anyone have any ideas what else can be added to avoid unnecessary records in the database and prevent its growth?

### DIFF
--- a/app/jobs/verify_emails_job.rb
+++ b/app/jobs/verify_emails_job.rb
@@ -24,7 +24,8 @@ class VerifyEmailsJob < ApplicationJob
     contact = Contact.find(contact_id)
     contact_ids = Contact.where(email: contact.email).where('created_at > ?', time).pluck(:id)
 
-    r = ValidationEvent.where(validation_eventable_id: contact_ids)
+    r = ValidationEvent.where(validation_eventable_id: contact_ids).order(created_at: :desc)
+    # return false if r[0].success == false
 
     r.present?
   end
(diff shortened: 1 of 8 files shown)


OlegPhenomenon:
> To avoid unnecessary records in the database maybe when we'll decide to destroy the contact we should delete all ValidationEvent records associated with the contact?

When a contact is invalid or the FD status is set for a domain, the contact is not deleted. But, there is another case that might fit your idea. There are invalid domains that do not belong to any domain, they are by themselves. Here they are removed within six months. But again, there is a legal side to this, should we keep track of the data to explain the reason for the deletion? This should be Timo sorted out. Also, if the registrar decides to delete the contact, then the information can also be removed from the ValidationEvent

OlegPhenomenon:
There is another idea: if the domain was validated successfully and its previous validation was also successful, then we can delete the previous record

PR #2204, opened 2021-10-29
added resolver to check ns record | Don't MERGE yet! 

### LINKED ISSUE #2202: Nameserver record validation
We are scanning the registry daily for cdnskey record changes (csync). This enables us to also detect invalid ns records in the registry.

we should register nameserver records that
* do not respond
* have no record about the domain it's queried about (is registered to server in the TLD registry)

when such a ns record is detected we should wait for at least 2 additional confirmations by the next scans (identically to csync scans). If three consecutive scans detect the same issue we should flag the record as invalid and notify tech contact via email and registrar via poll message about the issue.

If the issue is still there after a month - notify again

OlegPhenomenon:
Close #2202

### DIFF
--- a/app/jobs/nameserver_record_validation_job.rb
+++ b/app/jobs/nameserver_record_validation_job.rb
@@ -10,6 +10,7 @@ class NameserverRecordValidationJob < ApplicationJob
       end
     else
       result = validate(nameserver)
+
       check_validation_info(result: result, nameserver: nameserver)
     end
   end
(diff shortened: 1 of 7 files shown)


OlegPhenomenon:
Validation event table should be only for email validation. If we will add in this table nameserver validation results, we also will have problem with this model. Better solution to make separate validation model for nameservers and other thing

PR #2205, opened 2021-11-02
added trumail validator after update and implement simple email valid…

### LINKED ISSUE #2184: Mail validation after update
validate email address with turemail on new contact object creation and every email address update of a contact. same logic for multiple retries before taking action applies here as for regular validation checks. This is in addition to the regexp checks that should be done on any new and updated email record.

* contact/email address is validated only in case of success (no errors)
* validation is effective only for a specific validation level it has passed and levels below it (three validation levels: regexp, mx, smtp) - if an email address is validated with mx level validator the validation is good for both regexp and mx level validations, but when email validator is run on smtp level validations then that address should be considered as not validated.

if email address is changed the contact object must be revalidated

OlegPhenomenon:
close #2184

### DIFF
--- a/app/interactions/actions/contact_create.rb
+++ b/app/interactions/actions/contact_create.rb
@@ -12,7 +12,24 @@ module Actions
       maybe_remove_address
       maybe_attach_legal_doc
       validate_ident
+      maybe_change_email
       commit
+    end
+
+    def maybe_change_email
+      return if Rails.env.test?
+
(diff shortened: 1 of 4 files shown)


vohmar:
error 2005: "Parameter value syntax error: invalid email address"

vohmar:
* there is no valdation on object create.
* FD was not lifted after updating the invalid email address with a valid one

PR #2206, opened 2021-11-04
added validation before update

### LINKED ISSUE #1897: Add sub-zone dnskey check on adding new dnskey to a domain
On EPP/REPP domain update with adding new DNSKEY to domain object we should run the sub-zone check to make sure the key being added matches the key available in the authoritative nameservers. Delayed job should be added for new domain regsitrations and domain updates for configurable delay to check the validity of the trust-chain after the the change is made in registry.

if validation fails for all the host records then remove the key from the .ee zone

if validation fails for some of the hosts then the results should be configurable:
option 1: remove the host record from the .ee zone that does not have the valid dnskey record
option 2: remove the dnskey from .ee zone

Inform registrar via epp poll and domain contacts via email about the changes made to the registration

Key checking utility is available with the CSYNC functionality

Error message to return on failed check:

2308    "Data management policy violation: DNSKEY does not match or not found in the authoritative nameservers"

OlegPhenomenon:
Close #1897


vohmar:
Please add an option on running the validation job that would hold off notifying regsitrants and registrars about found issues so we can do the the initial run without alerting anyone and see what kind of issues we will find in production

Using these results we need to improve the notification polls and emails that we are sending out

PR #2219, opened 2021-11-19
Add linkpay

### LINKED ISSUE #2210: Upgrade to EveryPay APIv4
in november 2021 the support for v3 of the everypay API will end so all integrations should be upgraded to v4.

doc:
https://support.every-pay.com/downloads/everypay_apiv4_integration_documentation
https://support.every-pay.com/ms/integration/

v4 enables using linkpay and open banking features so we can drop using swedbank's and seb's banklinks and direct all payments to our primary bank account. WE can also add linkpay feature to all emails and pdf invoices.

DinSmol:
close #2210
Replaced banklinks with linkpay (everypay).
Added linkpay to email and pdf.
Added qr to pdf.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,6 +70,7 @@ gem 'jquery-ui-rails', '6.0.1'
 gem 'pdfkit'
 gem 'que'
 gem 'que-web'
+gem 'rqrcode', '~> 2.0'
 gem 'sidekiq'
 
 gem 'company_register', github: 'internetee/company_register',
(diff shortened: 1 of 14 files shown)


OlegPhenomenon (approved):

vohmar:
Few issues:
* replace the payment button in registrar with a button called pay - it shows currently card payment icon that can be confusing
* the payment windows should be openened in new tab or browser window so user can close it after the payment process is finished

now the important one
* even though the payment was successful the invoice was not marked as paid

PR #2228, opened 2021-11-29
Force delete notes to registrar

### LINKED ISSUE #2211: Force delete notes to registrar
We save ForceDelete notes in admin - ie email addresses if fd is set automatically by the email validator, comment by admin if force delete is set manually because of some policy violation, or in the future by business registry nr or phone nr validator

admin:
image

registrar:
image

DinSmol:
close #2211

### DIFF
--- a/app/views/epp/domains/info.xml.builder
+++ b/app/views/epp/domains/info.xml.builder
@@ -11,7 +11,7 @@ xml.epp_head do
         xml.tag!('domain:name', @domain.name)
         xml.tag!('domain:roid', @domain.roid)
         @domain.statuses.each do |s|
-          xml.tag!('domain:status', 's' => s)
+          xml.tag!('domain:status', 's' => s, 'n' => @domain.status_notes[s])
         end
 
         xml.tag!('domain:registrant', @domain.registrant.code)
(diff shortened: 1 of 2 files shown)


vohmar:
there is still the schema issue here.
After deploying i receive wrong path error when trying to access domain info view in registrar

this is the error returned by the epp:
  <response>
    <result code="2100">
      <msg lang="en">Wrong path</msg>
    </result>
    <trID>
      <clTRID>Märi Änn ED-1639493780</clTRID>
      <svTRID>ccReg-4676182991</svTRID>
    </trID>
  </response>
</epp>

and the request:
  <command>
    <info>
      <domain:info xmlns:domain="https://epp.tld.ee/schema/domain-ee-1.3.xsd">
        <domain:name>katki-mx-adm.ee</domain:name>
      </domain:info>
    </info>
    <clTRID>Märi Änn ED-1639493780</clTRID>
  </command>
</epp>

we need to support new and current schema files here

vohmar:
Looking at the RFC and schemas I think that changing the schema is not necessary for this and this option should be available out of the box:

<domain:status s="serverForceDelete">"admin pani asdfasd@adasdfa.adfasd"</domain:status>

Please correct me if I am wrong

PR #2230, opened 2021-11-29
Test result from one db to another

OlegPhenomenon:
Close #https://github.com/internetee/accreditation_center/issues/131


OlegPhenomenon:
Need to discuss another approach!

OlegPhenomenon:
Decided to use a different approach

PR #2232, opened 2021-12-07
ForceDelete prohibit settings serverDeleteProhibited

### LINKED ISSUE #2218: ForceDelete prohibits setting serverDeleteProhibited in admin
If ForceDelete is set for a domain setting serverDeleteProhibited status fails with 

image

steps to reproduce:
1. edit status
2. actions, schedule fd
3. actions, add new status
4. select delete prohibited and click save
5. error is returned, request fails

after cancelling fd the deleteProhibited appears as if it were set before setting FD

DinSmol:
close #2218

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Give test coverage reporter executable permissions
         run: chmod +x cc-test-reporter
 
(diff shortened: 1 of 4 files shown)


vohmar:
adding serverDeleteProhibited to a FD domains still fails
image

Issue #2253, opened 2021-12-27
Insufficient test coverage for domain registry


maricavor:
Add missing or additional integration tests for domain registry

PR #2255, opened 2021-12-27
Added missing integration/system tests for account_activities

maricavor:
Related to #2253


yulgolem:
If that's a PR, where is the description? Which issue does it close?

yulgolem:
And as for tests failing with nokogiri - we should either raise the ruby version or fix nokogiri version to address this issue first.

yulgolem (approved):

PR #2260, opened 2022-01-04
Added rubocop gem for development environment

maricavor:
Added rubocop gem for development environment


maricavor:
Will be used through IDE

yulgolem:
BTW I'll also suggest to plug in the Standard gem - https://github.com/testdouble/standard
Take a look at that.

PR #2263, opened 2022-01-07
Fix domain status history view

### LINKED ISSUE #834: Admin: domain history - 3 identical records created on accepting registrant change request
steps to reproduce:
* create a request to change registrant of a domain
* click on the link that was sent to current registrant and accept the change
* go to admin find the domain in the list of domains (/admin/domains) and click it open
* click on history button to open domain_history view (/admin/domains/86931/domain_versions)
* you will see 4 new records - 1. for the registrant change request and then 3 identical records for the accepting

expected result is 2 new records - for the request and for the actual change.

maricavor:
Related to #834

Fixed duplication of records in domain status history after status change by admin.

Added pagination for domain versions with per_page = 10
Added I18n pluralization for result_count

### DIFF
--- a/app/controllers/admin/domains_controller.rb
+++ b/app/controllers/admin/domains_controller.rb
@@ -63,8 +63,11 @@ module Admin
     # rubocop:enable Metrics/MethodLength
 
     def versions
+      per_page = 10
       @domain = Domain.where(id: params[:domain_id]).includes({ versions: :item }).first
       @versions = @domain.versions
+      @last_version = @versions.last
+      @old_versions = Kaminari.paginate_array(@versions.not_creates.reverse).page(params[:page]).per(per_page)
     end
 
     def keep
--- a/app/controllers/admin/domains_controller.rb
+++ b/app/controllers/admin/domains_controller.rb
@@ -68,7 +68,9 @@ module Admin
       @domain = Domain.where(id: params[:domain_id]).includes({ versions: :item }).first
       @versions = @domain.versions
       @last_version = @versions.last
-      @old_versions = Kaminari.paginate_array(@versions.not_creates.reverse).page(params[:page]).per(DEFAULT_VERSIONS_PER_PAGE)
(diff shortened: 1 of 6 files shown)


app/controllers/admin/domains_controller.rb
```diff
@@ -63,8 +63,11 @@ def update
     # rubocop:enable Metrics/MethodLength
 
     def versions
+      per_page = 10
```
yulgolem, line 66: Looks like some magic number to me. Consider using a constant like DEFAULT_RECORDS_PER_PAGE and defining it in the controller itself.

maricavor: Thanks! Just saw the same apporach in other controllers. Actually I wanted to add paginates_per 10 to the model itself, but did not want to make it general rule. I will do like you suggested - add constant in the controller!

app/controllers/admin/domains_controller.rb
```diff
@@ -65,6 +67,8 @@ def update
     def versions
       @domain = Domain.where(id: params[:domain_id]).includes({ versions: :item }).first
       @versions = @domain.versions
+      @last_version = @versions.last
+      @old_versions = Kaminari.paginate_array(@versions.not_creates.reverse).page(params[:page]).per(DEFAULT_VERSIONS_PER_PAGE)
```
yulgolem: Do our user have any ability to input a number of entries per page? If so I'd suggest to use something like 
params[:per_page] || DEFAULT_VERSIONS_PER_PAGE , maybe moved to some helper method.
If not- ok then, we are good to go

yulgolem (approved):

app/controllers/admin/domains_controller.rb
maricavor: No, there is no per page input, not necessary

OlegPhenomenon (approved):

DinSmol (approved):

DinSmol (approved):

PR #2271, opened 2022-01-13
Update FD notes when basis of the FD is changed

### LINKED ISSUE #2216: Update FD notes when basis of the FD is changed/updated
example:
* FDis set to a domain with invalid email address
* some time later another email address bounces and is added to the bounce list - here the FD notes should be updated with another email address
* later the first email address that initiated the FD is replaced with a valid one - FD notes update to remove the email address, FD remains as another domain is still in bounce list

sources for FD initiation:
* manual - set by administrator (ie non-existent priv registrant/contact; : this can also be removed only manually  (requires a specific note - reason and admin; removal will be automated later)
* email bounce
* email validator (regex, mx, smtp)
* business registry validation
* citizenship registry validation (not yet implemented as we do not have API access to the citizens registry)

DinSmol:
close #2216

### DIFF
--- a/test/models/domain/force_delete_test.rb
+++ b/test/models/domain/force_delete_test.rb
@@ -443,7 +443,7 @@ class ForceDeleteTest < ActionMailer::TestCase
 
     domain.reload
     contact_first.update(
-      email: 'correct_email@internet2.ee',
+      email: 'aaa@bbb.com',
       email_history: email_1
     )
 
--- a/test/models/domain/force_delete_test.rb
+++ b/test/models/domain/force_delete_test.rb
@@ -411,55 +411,58 @@ class ForceDeleteTest < ActionMailer::TestCase
     assert notification.text.include? asserted_text
   end
 
-  def test_force_delete_notes
+  def test_add_invalid_email_to_domain_status_notes
     domain = domains(:airport)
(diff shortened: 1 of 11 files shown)


test/models/domain/force_delete_test.rb
```diff
@@ -411,6 +411,59 @@ def test_schedules_force_delete_invalid_contact
...
+    assert domain.force_delete_scheduled?
+    assert_equal 'invalid_email', domain.template_name
+    assert_equal Date.parse('2010-09-19'), domain.force_delete_date.to_date
+    assert_equal Date.parse('2010-08-05'), domain.force_delete_start.to_date
+
+    assert_equal domain.status_notes[DomainStatus::FORCE_DELETE], email_2
+    notification = domain.registrar.notifications.last
+    assert notification.text.include? asserted_text
+  end
+
```
OlegPhenomenon, line 466: maybe it makes sense to break the test into parts and transfer some logic to separate modules? it's very hard to read this, it's more like not a unit test, but a system test. in addition, calling the update_attribute method must trigger some action on the part of the user/system. it makes sense to write this test in a job / task / system test, depending on where the FD records are updated and where is the functionality that you wrote applied

OlegPhenomenon (changes requested):

OlegPhenomenon (approved):

PR #2294, opened 2022-02-22
Removing expired records from the validations table

### LINKED ISSUE #2236: Remove expired records from the validations table
successfully validated email records are valid for 1 year after that period of time the email address should be revalidated. On revalidating the record old expired records should be removed from the validations table.

this could be done by either the same validation method when expired validation is detected or by a separate method

Test:
* run email validation at mx validation level
* choose any successfully validated record for further monitoring
* run email validation at mx level - check that there is no now validation record created for the chosen email address
* change validity period setting to 1 minute
* run email validation at mx level - check that there is now new validation record for the chosen email and the old record is deleted

DinSmol:
close #2236


maricavor (approved):

vohmar:
run  verify_email 3 times to make sure there are not new records created

updated create_date of all records in validation_events table to '2021-03-21' and run rake verify_email:check_all -- --check_level=mx'

as a results
182216 records with create_date '2021-03-21' still in the table out of which
13766 have success value 'true'

after three consecutive runs of verify_email there are no new records created and there is
105 records with create_date > '2022-03-24' (63 distinct)
66 records with 'true' (50 distinct)

so most true records with create date more than 1 year old were not deleted and revalidated, not sure why.

OlegPhenomenon (approved):

OlegPhenomenon (approved):

maricavor (approved):

PR #2298, opened 2022-02-28
dnssec ans ns validation message

### LINKED ISSUE #2296: NS and dnskey validation error messages
NS validation:
No any answer comes from **ns.almic.eu**. Nameserver not exist
correct: DNS Server **ns.almic.eu** Not Responding

OlegPhenomenon:
Close #2296


vohmar:
midified couple of messages
resolve conflict

PR #2308, opened 2022-03-10
Admin: fix CSV data output

### LINKED ISSUE #2303: Admin: CSV files should include at least the same set of data available in admin UI
Related to https://github.com/internetee/registry/issues/2275

concerns csv download options in
/admin/domains
/admin/contacts
/admin/invoices
/admin/accounts

CSV download just puts the content of respective main db table into the file so for domains csv its the content of domains table. Problems

* the table used often does not include all the data expected ie domains table includes the object id  of registrant and registrar but not their names and contact data
* the table includes db internal ids that are useless outside of the database 

expected result:
* csv should include all the same data that is returned to the users in admin ui
* additional data can be added in csv, but after the main data and it should not include any data table specific data like row id's

thiagoyoussef:
close #2303


thiagoyoussef:
Tests are failing because I haven't finished the implementation

app/helpers/application_helper.rb
```diff
@@ -9,30 +9,6 @@ def env_style
     "background-image: url(#{image_path("#{unstable_env}.png")});"
   end
 
-  def ident_for(contact)
```
thiagoyoussef, line 12: I moved this method to `app/models/contact.rb` and changed the name to `ident_human_description`

OlegPhenomenon (approved):

app/models/invoice.rb
```diff
@@ -128,6 +145,16 @@ def self.create_from_transaction!(transaction)
 
   private
 
+  def receipt_date_status
```
thiagoyoussef, line 148: Is the method name clear?

app/models/account.rb
```diff
@@ -12,4 +12,12 @@ class Account < ApplicationRecord
   def activities
     account_activities
   end
+
+  def as_csv_row
+    [id, balance, currency, registrar]
+  end
+
+  def self.csv_header
+    ['Id', 'Balance', 'Currency', 'Registrar'].freeze
```
OlegPhenomenon, line 21: If you decide to freeze this value and use it as global constant, then you should to move it into the top of file and assigned it like constant `CSV_HEADER = ['Id', 'Balance', 'Currency', 'Registrar'].freeze`, otherwise freeze method not make sense to add. So there are two option: first - move to top and make this value as constant, second - remove freeze method

app/models/contact.rb
```diff
@@ -569,4 +569,27 @@ def registrant?
...
+      email,
+      created_at.to_formatted_s(:db),
+      registrar,
+      phone,
+    ]
+  end
+
+  def self.csv_header
+    ['Name', 'ID', 'Ident', 'E-mail', 'Created at', 'Registrar', 'Phone'].freeze
+  end
```
OlegPhenomenon, line 594: Here is the same, you has identical principles in as_csv_row method and in csv_header, but csv_header looks like you tried to put there constant value. Why?

app/models/domain.rb
```diff
@@ -753,6 +753,35 @@ def contact_emails_verification_failed
...
+
+    contact = ObjectVersionsParser.new(ver).parse if contact.nil? && ver
+
+    contact.try(:name) || 'Deleted'
+  end
+
+  def self.csv_header
+    ['Domain', 'Registrant', 'Valid to', 'Registrar', 'Created at', 'Statuses', 'Contacts code', 'Force delete date', 'Force delete data'].freeze
+  end
+
```
OlegPhenomenon, line 784: The same

app/models/invoice.rb
```diff
@@ -117,6 +117,23 @@ def e_invoice_sent?
...
+      total,
+      currency,
+      seller_name
+    ]
+  end
+
+  def self.csv_header
+    ['Number', 'Buyer', 'Due Date', 'Receipt Date', 'Issue Date', 'Total', 'Currency', 'Seller Name'].freeze
+  end
+
```
OlegPhenomenon, line 136: The same

app/models/version/contact_version.rb
```diff
@@ -4,4 +4,21 @@ class Version::ContactVersion < PaperTrail::Version
...
+      contact.ident_human_description,
+      contact.registrar,
+      event,
+      created_at.to_formatted_s(:db)
+    ]
+  end
+
+  def self.csv_header
+    ['Name', 'ID', 'Ident', 'Registrar', 'Action', 'Created at'].freeze
+  end
```
OlegPhenomenon, line 23: again, lol :D

app/models/version/domain_version.rb
```diff
@@ -43,4 +55,8 @@ def self.contact_unlinked_more_than?(contact_id:, period:)
 
     count_by_sql(sql).nonzero?
   end
+
+  def self.csv_header
+    ['Name', 'Registrant', 'Registrar', 'Action', 'Created at'].freeze
+  end
```
OlegPhenomenon, line 61: or noooo, not again :D

OlegPhenomenon (changes requested):

app/models/account.rb
thiagoyoussef: That's true! I will remove the freeze method ;)

PR #2309, opened 2022-03-11
business registry contacts - endpoints for send status of contacts and receive confirmation

OlegPhenomenon:


app/models/registrant_user.rb
```diff
@@ -20,16 +20,34 @@ def country
...
-    return []
+    []
+  end
+
+  def do_need_update_contact?
+    return { result: false, counter: 0 } if companies.blank?
+
+    counter = 0
+    companies.each do |company|
+      contacts = Contact.where(ident: company.registration_number, ident_country_code: 'EE')
```
maricavor, line 34: its better to use one Active Record query here, than iterations:
`counter  += Contact.where(ident: company.registration_number, ident_country_code: 'EE')
                                   .where('name != ?', company.company_name).size`

maricavor (commented):
Is structure.sql ok?

maricavor (approved):

PR #2315, opened 2022-03-21
Fixed notifications about automatic contact name update 

### LINKED ISSUE #2307: Registrant API: move notifications about automatic contact name updates to bulk change poll messages
currently we send a poll message out to registrars one for each contact object that is changed. If more than 100 is update at a time this would potentially create too many messages so that processing those automatically will instantly fill the registrars' query rate limit

In case of registry lock we already user epp change poll extension for machine readable poll messages.

Consider implementing change polls for registrant automated contact updates with multiple updates (ie 100) defined in a single message

rfc for change poll: https://www.rfc-editor.org/rfc/rfc8590.html

not entirely sure if the rfc allows putting changes of multiple objects into one change poll message. Alternatively it's possible to just add more info into the standard unstructured poll messages

maricavor:
Close #2307 
Moved notifications about automatic contact name update to bulk change poll message:


app/models/action.rb
```diff
@@ -2,18 +2,40 @@ class Action < ApplicationRecord
...
+  def bulk_action?
+    !!subactions.exists?
+  end
+
+  def to_non_available_contact_codes
+    return [] unless bulk_action?
+
+    subactions.map do |a|
+      {
+        code: a.contact&.code,
```
OlegPhenomenon, line 35: but if a.contact is nil, then code will equal nil. What is the point of element of subactions list if it will store some element with nil value? Are you sure is it need this logic here?

app/views/epp/poll/_action.xml.builder
```diff
@@ -1,9 +1,10 @@
 builder.extension do
   builder.tag!('changePoll:changeData',
-               'xmlns:changePoll' => Xsd::Schema.filename(for_prefix: 'changePoll')) do
+               'xmlns:changePoll' => Xsd::Schema.filename(for_prefix: 'changePoll', for_version: '1.0')) do
```
OlegPhenomenon, line 3: In some files, including this one, it is not quite in terms of clean code declaration code. Since with each schema change we need to change the schema version as well, it can take some time to refactor code like this.

For example, if you look in the `app/views/epp/domain/create.xml.builder` file, there is a class variable in the versions in the place where the version and prefix are specified. I suggest you do the same here

db/structure.sql
```diff
@@ -826,8 +827,7 @@ CREATE TABLE public.dnskeys (
     updator_str character varying,
     legacy_domain_id integer,
     updated_at timestamp without time zone,
-    validation_datetime timestamp without time zone,
-    failed_validation_reason character varying
```
OlegPhenomenon, line 830: why was it removed?

OlegPhenomenon (changes requested):

app/models/action.rb
maricavor: because contact is now optional, but in this context, yes, it cannot be nil, because we need code attribute

db/structure.sql
maricavor: I have no idea

app/views/epp/poll/_action.xml.builder
maricavor: in case of domain requests it is possible to parse schema prefix and version, domain_controller uses method parse_schemas_prefix_and_version and assigns: @schema_version = res[:version]
@schema_prefix = res[:prefix], which then are used in views. In poll requests there is no changePoll version.

thiagoyoussef (approved):

vohmar:
error on deploy:


```
fatal: [staging.internet.ee]: FAILED! => {
    "changed": false,
    "cmd": "/home/registry/.rbenv/shims/bundle install",
    "invocation": {
        "module_args": {
            "binstub_directory": null,
            "chdir": "/home/registry/deploy/releases/20220331160227",
            "clean": false,
            "deployment_mode": false,
            "exclude_groups": null,
            "executable": null,
            "extra_args": null,
            "gem_path": null,
            "gemfile": null,
            "local": false,
            "state": "present",
            "user_install": true
        }
    },
    "msg": "Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_check                                                                                                                                                                                        er)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spel                                                                                                                                                                                        l_checker)' instead.\ngoogle-protobuf-3.19.4-x86_64-linux requires ruby version                                                                                                                                                                                         < 3.1.dev, >= 2.5,\nwhich is incompatible with the current version, ruby 3.1.1p1                                                                                                                                                                                        8",
    "rc": 5,
    "stderr": "Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_ch                                                                                                                                                                                        ecker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, s                                                                                                                                                                                        pell_checker)' instead.\ngoogle-protobuf-3.19.4-x86_64-linux requires ruby versi                                                                                                                                                                                        on < 3.1.dev, >= 2.5,\nwhich is incompatible with the current version, ruby 3.1.                                                                                                                                                                                        1p18\n",
    "stderr_lines": [
        "Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)                                                                                                                                                                                        ' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_c                                                                                                                                                                                        hecker)' instead.",
        "google-protobuf-3.19.4-x86_64-linux requires ruby version < 3.1.dev, >=                                                                                                                                                                                         2.5,",
        "which is incompatible with the current version, ruby 3.1.1p18"
    ],
    "stdout": "Fetching gem metadata from https://rubygems.org/..........\n",
    "stdout_lines": [
        "Fetching gem metadata from https://rubygems.org/.........."
```

app/views/epp/poll/_action.xml.builder
OlegPhenomenon: Theoretically, a version may appear in the future. In any case, hardcoding versions in plain text is not considered the best approach. There must be at least a constant in the file that specifies the schema and prefix versions. It was the same with domains like you have now. There were no new schemas, but then the prefix and version changed and I had to go through a large number of methods, classes and tests and change versions and prefixes everywhere. This needs to be done so that if in the future it will be necessary to change something, so that it does not take much time.

Since your method accepts a prefix and a version, it is obvious that these are dependent things and they are taken out as a separate constant or raised up in the parent class.

OlegPhenomenon (approved):

PR #2323, opened 2022-03-24
update docker files 24-04-2022

OlegPhenomenon:
~~This is just a test. I am not sure that it works as expected, because some problems with smtp connection. I need figure it out, if it the dockerfile reason or not.~~ Any way, this docker image reduced the size of image from  2.77GB to 1.62GB!

UPDATED: I remove couple based images like ruby-slim-buster and ruby-alpline, remove registry image and rebuild again and problem with smtp is gone. So, seems everything works as expected

1. Notes: this is dockerfile only for development environment. For Kubernetes staging and production use another instances
2. Notes: this dockerfile use alpine linux distribute, so it doesn't contain bash terminal. So, to run this docker file in docker-compose.yml need to change `command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails server -b 0.0.0.0"` to `command: sh -c "rm -f tmp/pids/server.pid && bundle exec rails server -b 0.0.0.0"` in registry section and the same for epp section.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,223 @@
+.git
+.gitignore
+
+# Created by https://www.gitignore.io/api/git,ruby,rails,jetbrains+all
+# Edit at https://www.gitignore.io/?templates=git,ruby,rails,jetbrains+all
+
+### Git ###
+# Created by git for backups. To disable backups in Git:
+# $ git config --global mergetool.keepBackup false
+*.orig
+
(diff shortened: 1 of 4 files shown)


maricavor (approved):

thiagoyoussef (approved):

OlegPhenomenon:
not actual yet

PR #2339, opened 2022-04-06
added parallel functionality for running tests

OlegPhenomenon:
Added a new feature in test setup that allows you to run tests in parallel

My test execution accelerated was 1:53 and then 1:12 
For 40 seconds in my case, the execution of the tests accelerated. When you have time, please try it too and write down your performance to be sure that these changes can increase the speed of running tests. Thanks


thiagoyoussef (commented):
The performance was the same on both cases for me, 16 minutes!
Maybe it's an exception because of my chromedrive problem

PR #2341, opened 2022-04-08
added auction list to admin panel

OlegPhenomenon:
Not for merge yet.
Need to add tests


vohmar:
looking good. Few thought and issues:
* please move the auctions selection in the settings menu down to system section between zones and blocked domains
* the default view shows only auctions with no_bids even though status selection is empty (maybe because the whole list is not returned and the list starts with records with no_bids status)
* pagination is missing
* Reset btn rename to reset or reset filter; button tooltips missing ("translation missing")
* csv upload is not working - internal error. Maybe because of invalid csv format (some tip or hint should be there somewhere to inform user on what the proper csv import file looks like) 
* created_ at from and created at until and results per page options do not work

vohmar:
looking even better

* registration code tends to extend the column so that other like create date are wrapped into more lines making the whole table look bad - maybe its better to fix the width of reg code field and provide some copy help for the content
image
image

* Type automatically should be replaced with automatic or auto; manually with manual
* trying to upload csv for adding new auctions - getting "invalid csv format" error. What is correct format? We should add some help about this in the UI
* when adding an auction from reserved list to auction there should be a warning to let user know that it is reserved and as a result this domain will be removed from the reserved list
* blocked domains should not be possible to add to the auction
* registered domains should not be possible to add to the auction
* domains in active disputed list should not be possible to add to the auction
* only domains under the zones served by the registry should be possible to add to the auction and only the first level ie domeen.ee or domeen.pri.ee - was able to add domains like aaaa.e, aaaa, aaa.aaa.aaa
* is it possible to refresh the list automatically when new domain(s) is/are added to the list ie my list was sorted by created at date descending, but adding newly added domains did not automatically appear on top of the list , had to reorder the list to see if and what was added.

--

import from reserved list is crazy good, buttons work csv download works - nicely done!

vohmar:
Getting close to amazing. Few things still though:
* sorting by type does not work - auto always at the top, manual at the end
* still missing warnings when adding domains from reserved list to auction

PR #2344, opened 2022-04-11
fix poll message spam after validation email

### LINKED ISSUE #2343: Poll messages sent on every failed email validation job scan?
It look like registry is generating an sending poll messages to registrants whenever new failed record is added to the validation events table.

In reality the poll message should only be generated when the actual status changes for a domain  ie ForceDelete is set

OlegPhenomenon:
Close #2343

### DIFF
--- a/app/interactions/domains/force_delete/notify_registrar.rb
+++ b/app/interactions/domains/force_delete/notify_registrar.rb
@@ -6,18 +6,26 @@ module Domains
       end
 
       def notify_without_email
-        domain.registrar.notifications.create!(text: I18n.t('force_delete_set_on_domain',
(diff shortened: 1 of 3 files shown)


vohmar:
when the force delete is set 3 poll messages are sent to the registrar - should be only one

vohmar:
one poll message is sent per object-domain relation. So it the contact is in all three roles - 3 polls are sent, if in any 2 then only 2 messages. Correct result would be a single poll message per domain that got force delete applied

vohmar:
works like a charm, thank you @OlegPhenomenon. Will merge it as soon as review has also approved

maricavor (approved):

Issue #2364, opened 2022-04-22
Email validation job cannot cope with production load


vohmar:
jobs in sidekiq que stay there for more than 20 minutes, CPU running at 100%, no load on DB
Makes millions of queries to database like this:

```
Apr 22 11:40:52 reg-l-epp adm.internet.ee[10710]: [ActiveJob] [VerifyEmailsJob] [65a992fd-9d1f-4f24-aa48-cfd2880560d8]   Domain Load (15.9ms)  SELECT "domains"."id", "domains"."name", "domains"."registrar_id", "domains"."valid_to", "domains"."registrant_id", "domains"."transfer_code", "domains"."created_at", "domains
"."updated_at", "domains"."name_dirty", "domains"."name_puny", "domains"."period", "domains"."period_unit", "domains"."creator_str", "domains"."updator_str", "domains"."outzone_at", "domains"."delete_date", "domains"."registrant_verification_asked_at", "domains"."registrant_verification_token", "domains"."pending_jso
n", "domains"."force_delete_date", "domains"."statuses", "domains"."status_notes", "domains"."statuses_before_force_delete", "domains"."upid", "domains"."up_date", "domains"."uuid", "domains"."locked_by_registrant_at", "domains"."force_delete_start", "domains"."force_delete_data", "domains"."json_statuses_history" FR
OM "domains" WHERE "domains"."id" = $1 LIMIT $2  [["id", 170193], ["LIMIT", 1]]
```

PR #2365, opened 2022-04-22
Modified verify email rake task

maricavor:
Related to #2364

Modified verify email rake task so the contacts batch size and limit can be added as options: 
`bundle exec rake verify_email:check_all -- --batch_size=1000 --limit=100`
or
`bundle exec rake verify_email:check_all -- -b1000 -l100`

By default batch_size is 10000 and limit is 0 (0 = unlimited).

NB! Changing batch_size will not stop the task from finding all the contacts, which need email validation. VerifyEmailsJob will start for each contact. Adjusting limit option will limit the amount of contacts that need to be validated, thus helping to protect from overloading by sidekiq jobs and to troubleshoot task problems in production environment.

### DIFF
--- a/lib/tasks/verify_email.rake
+++ b/lib/tasks/verify_email.rake
@@ -13,6 +13,7 @@ namespace :verify_email do
       domain_name: nil,
       check_level: 'mx',
       spam_protect: false,
+      batch_size: 10_000,
     }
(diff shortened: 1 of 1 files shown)


OlegPhenomenon:
PR can be close. The issue was different

PR #2367, opened 2022-04-25
added restriction for parallel record data in db for the same objects

OlegPhenomenon:


app/interactions/actions/email_check.rb
```diff
@@ -62,9 +62,13 @@ def save_result(result)
                                              type: 'AAAA')
 
         result_validation.present? ? result.success = true : result.success = false
-        validation_eventable.validation_events.create(validation_event_attrs(result))
+        ActiveRecord::Base.connection_pool.with_connection do
```
maricavor, line 65: One problem of ActiveRecord::Base.connection_pool.with_connection is that it always prepare a connection even if the code inside does not need it. So better use: 
```
Thread.new do
  begin
    # Do whatever
  ensure
    ActiveRecord::Base.connection_pool.release_connection
  end
end
```
Check [this](https://stackoverflow.com/questions/23596183/how-to-manage-opening-and-closing-database-connections-while-working-with-active/26733385)

OlegPhenomenon: I'll take note. Thanks

OlegPhenomenon:
PR is close. The issue was different #2368

PR #2369, opened 2022-04-27
Add validation event check force delete to a job

thiagoyoussef:


app/jobs/validation_event_check_force_delete_job.rb
```diff
@@ -0,0 +1,40 @@
+class ValidationEventCheckForceDeleteJob < ApplicationJob
```
thiagoyoussef, line 1: change the name

lib/tasks/check_force_delete.rake
```diff
@@ -0,0 +1,11 @@
+desc 'Check Force Delete'
+task check_force_delete: :environment do
+  invalid_contacts = Contact.joins(:validation_events).select do |contact|
+    events = contact.validation_events
+    events.mx.count >= 3 || events.regex.present?
```
thiagoyoussef, line 5: Use `MX_CHECK`

thiagoyoussef: invalid

lib/tasks/check_force_delete.rake
```diff
@@ -0,0 +1,11 @@
+desc 'Check Force Delete'
+task check_force_delete: :environment do
```
thiagoyoussef, line 2: Add args to use batches and schedule delay between jobs

OlegPhenomenon (approved):

vohmar:
To run the job you need to execute the rake task check_force_delete
3 optional arguments to call the task:

- `batch` - value can be true or false - default value is false - true: executes the job in batches, one job = many contacts - false: one job = one contact - this way Martin can see which one performs better
- `batch_size` - number of contacts that will be performed in each job - default 1000
- `batches_delay` - minutes between each job that will be performed - default 15 minutes

It's possible to call the task like this:

-  `rake check_force_delete` - will perform without batch, one contact = one job
-  `rake 'check_force_delete[true, 10000]'` - will perform with batch, 10000 contacts = one job, with 15 minutes of difference between jobs
-  `rake 'check_force_delete[true, 10000, 30]'` - will perform with batch, 10000 contacts = one job, with 30 minutes of difference between jobs

PR #2375, opened 2022-05-11
Prevent deprecated validation events

### LINKED ISSUE #2374: Refactor: prevent deprecated validation events in database
There are several validation events objects on the database that should have been removed
To prevent this to happen, add to destroy validation events when related contact is destroyed and contact email is changed

thiagoyoussef:
close #2374

### DIFF
--- a/app/interactions/actions/email_check.rb
+++ b/app/interactions/actions/email_check.rb
@@ -26,33 +26,27 @@ module Actions
       Rails.env.test? && check_level == 'smtp' ? :mx : check_level.to_sym
     end
 
+    def destroy_old_validations(validation_events, minimum_size)
+      return unless validation_events.count > minimum_size
+
+      validation_events.order!(created_at: :asc)
+      validation_events.first.destroy while validation_events.count > minimum_size
+    end
+
     def filtering_old_failed_records(result)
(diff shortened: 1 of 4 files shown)


app/interactions/actions/email_check.rb
```diff
@@ -26,33 +26,27 @@ def calculate_check_level
...
-      end
+      events = validation_eventable.validation_events
 
-      if @check_level == "mx" && result.success && validation_eventable.validation_events.count > 1
-        validation_eventable.validation_events.order!(created_at: :asc)
-        while validation_eventable.validation_events.count > 1
-          validation_eventable.validation_events.first.destroy
-        end
-      end
+      destroy_old_validations(events, 3) if @check_level == 'mx' && !result.success
```
OlegPhenomenon, line 39: What happen if we decide to change validation events limit? For example from 3 to 4. For this it will need to search code in whole project what responsibility for it. Also, if I didn't know nothing about validation, number 3 will be for me incomprehensible attribute. Please check validation_event model, there is should be constant where is assigned default validation limit and please change number 3 to this constant

(and yes, I know that I am too used numbers instead constants and this is also my fall :) )

OlegPhenomenon (changes requested):

PR #2382, opened 2022-05-24
Remove deprecated `statuses_before_force_delete` field

### LINKED ISSUE #2363: Reapplying statuses on ForceDelete removal can end up with invalid expired status
there are two similarly named columns in the domains table
statuses_before_force_delete
admin_store_statuses_history element in the json_statuses_history jsonb column

when is admin_store_statuses_history saved and updated?
consider combining the two elements into one

There was an issue with 4 domains so far that had expired status saved in admin_store_statuses_history and ok status saved in the statuses_before_force_delete column. When FD was removed system restored expired status for the domain instead of ok even though the valid_to date was 6 months away.

thiagoyoussef:
close #2363


db/migrate/20220524130709_remove_statuses_before_force_delete_from_domains.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveStatusesBeforeForceDeleteFromDomains < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :domains, :statuses_before_force_delete
```
OlegPhenomenon, line 3: Please google one method. I don't remember the name, but I remember exactly what the method was. The idea is that this method checks whether the given column exists or not. If it doesn't exist, it skips the deletion and if it does, it deletes

OlegPhenomenon (changes requested):

db/migrate/20220524130709_remove_statuses_before_force_delete_from_domains.rb
thiagoyoussef: `column_exists?`

OlegPhenomenon: yep

thiagoyoussef:
Staging test is failing because github packages is not working https://www.githubstatus.com/

OlegPhenomenon:
> Staging test is failing because github packages is not working https://www.githubstatus.com/

I was already thinking of asking Martin what the problem was with the deploying on the staging :D

PR #2400, opened 2022-07-11
Disclose additional registrant data

thiagoyoussef:
related https://github.com/internetee/registrant_center/issues/109

### DIFF
--- a/app/controllers/api/v1/registrant/contacts_controller.rb
+++ b/app/controllers/api/v1/registrant/contacts_controller.rb
@@ -62,7 +62,7 @@ module Api
           disclosed_attributes = reparsed_request_json[:disclosed_attributes]
 
           if disclosed_attributes
-            if disclosed_attributes.present? && contact.org?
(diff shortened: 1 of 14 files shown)


app/controllers/api/v1/registrant/contacts_controller.rb
```diff
@@ -62,7 +62,7 @@ def update
           disclosed_attributes = reparsed_request_json[:disclosed_attributes]
```
OlegPhenomenon, line 63: very huge action comes here and there is a lot of checkers. I think it will be better to devide this action to separate methods and give them good names to understand what here happens

app/models/concerns/zone/whois_queryable.rb
```diff
@@ -47,7 +47,7 @@ def registrar_vars
   def registrant_vars
```
OlegPhenomenon, line 47: Break each property on a new line here to make it easier to read this code, since there are two properties per line. Thanks

OlegPhenomenon (changes requested):

OlegPhenomenon (approved):

PR #2403, opened 2022-07-21
added exception for auctions with no-bids and registred-domains statuses

OlegPhenomenon:


vohmar:
works as expected. what about tests?

OlegPhenomenon:
> works as expected. what about tests?

Done

thiagoyoussef (approved):

PR #2407, opened 2022-07-29
extended csv domain export

### LINKED ISSUE #2359: CSV and PDF download puts only partial list in the file
in case of longer domain list (more than 10) only part of the whole list is loaded when user logs into the apartment and when user tries to download the list only that portion of domains is saved into the file.

Its not a pagination issue as with for example in took a list of 8 domains, set the page size to 6 records then still the whole list is put into the file as the whole list is actually loaded already. 

We need to put all the records that correspond to the active filter settings set by user in the domain list view.

The reason for pagination are users with long list of domains so maybe the file download will need to be put into a background job and provide user a download option once the file is generated?

OlegPhenomenon:
Close #2359

### DIFF
--- a/app/controllers/registrar/domains_controller.rb
+++ b/app/controllers/registrar/domains_controller.rb
@@ -36,14 +36,12 @@ class Registrar
         end
       end
 
-      @domains = @domains.per(params[:results_per_page]) if params[:results_per_page].to_i.positive?
-
       respond_to do |format|
         format.html
         format.csv do
           domain_presenters = []
 
-          @domains.find_each do |domain|
+          @q.result.find_each do |domain|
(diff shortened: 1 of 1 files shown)


OlegPhenomenon:
@vohmar 
exporting a CSV file no longer depends on pagination, but regarding the implementation of an asynchronous job for this purpose, there are a number of problems, you can read it here: https://stackoverflow.com/a/27084533

I see two solutions so far:
the first is to send the file for download via websockets and redis when the async task stops compiling a large list of domains. Since we will soon stop serving this portal, it seems to me that this will be a waste of time, and I hope everyone in the team will agree with me.

The second option is more compromise. If the number of domains is more than a certain number and it will really take a lot of time to generate a csv file, then one of the solutions for such a case is to send a csv file by mail

third option - your option?

PR #2411, opened 2022-08-03
added cache aproach to registrant api response for domains

OlegPhenomenon:


ratM1n:
First time loading took looong time, but eventually completed. Count shows that i have 40847 of 40847 domains (tech contact filter is off). But it will show me only 6 pages from this 40847 domains (24 domains per page), after that, "next" button goes inactive.

OlegPhenomenon:
> First time loading took looong time, but eventually completed. Count shows that i have 40847 of 40847 domains (tech contact filter is off). But it will show me only 6 pages from this 40847 domains (24 domains per page), after that, "next" button goes inactive.

it is quite expected that the first run will be very long, while the next runs should be fast. the goal was exactly that the NEXT runs should be fast. in any case, if not all domains are loaded into memory, then something is wrong.

PR #2412, opened 2022-08-05
added status sender for manually bind invoice for admin

OlegPhenomenon:


vohmar:
* when deployed on epp/repp/api machine, user cannot access invoices in registrar any more:
image

* would it be possible to add some kind of note/tag in billing for the source of the payment. So in this case by registry admin for example

PR #2418, opened 2022-08-14
Fix check force delete lift bug

### LINKED ISSUE #2413: ForceDelete lift task fails to lift
We have two reports from production about ForceDelete not being removed after the faulty email was corrected. In both cases the contact object was updated with correct email address

We were also able to reproduce the issue in test env where passing validation record was created for an email address listed as a cause for a ForceDelete on a domain.  Running lift task did not remove FD.

No errors from log to share at this time. 

to replicate:
1. set invalid email address to a registrant of a domain
2. run email validation job to set ForceDelete status to the domain
3. update the invalid registrant contact object with a valid email address
4. run ForceDelete_lift job - check if FD was removed from the domain

to replicate 2:
1. set non-existing email address to an admin contact of a domain
2. fix the email address by adding success=true record for the email in email_address_verifications table
3. run CheckForceDeleteLift job - check if FD was removed from the domain

thiagoyoussef:
close #2413

### DIFF
--- a/lib/tasks/check_force_delete.rake
+++ b/lib/tasks/check_force_delete.rake
@@ -1,7 +1,7 @@
 desc 'Check Force Delete'
 task check_force_delete: :environment do
   validations = ValidationEvent.failed.where(validation_eventable_type: 'Contact').uniq(&:validation_eventable_id)
-  return unless validations.present?
+  return if validations.blank?
 
   invalid_contact_ids = validations.select do |validation|
     contact = validation.validation_eventable


OlegPhenomenon (approved):

lib/tasks/check_force_delete.rake
```diff
@@ -3,8 +3,7 @@ task check_force_delete: :environment do
   validations = ValidationEvent.failed.where(validation_eventable_type: 'Contact').uniq(&:validation_eventable_id)
 
   invalid_contact_ids = validations.select do |validation|
-    contact = validation.validation_eventable
-    contact.need_to_start_force_delete? || contact.need_to_lift_force_delete?
+    validation.validation_eventable.need_to_start_force_delete?
```
OlegPhenomenon, line 6: where did you hide the check for need_to_lift_force_delete? how will he understand that force delete can be removed on a specific domain?

OlegPhenomenon (changes requested):

app/models/validation_event.rb
```diff
@@ -35,6 +35,8 @@ class ValidationEvent < ApplicationRecord
   scope :smtp, -> { where('event_data @> ?', { 'check_level': 'smtp' }.to_json) }
   scope :by_object, ->(object) { where(validation_eventable: object) }
 
+  after_create :check_force_delete_lift
+
```
OlegPhenomenon, line 39: I think that is a bad idea.

The last time we put checks for force delete and lift force delete into callbacks, firstly there were performance issues, and secondly, unexpected behavior happened when lift force delete worked in a situation where it was not supposed to.

In addition, there is an opinion that using callback is not a good practice in terms of support, code testing. But that's just an opinion, everyone has their own opinion :)

OlegPhenomenon (changes requested):

OlegPhenomenon (approved):

vohmar:
an issue detected - if user has the same invalid email with multiple contact objects related to a single domain registration the amount of false validation_event records is unlimited (seemingly) instead of three records. This is not a rare case in production

PR #2424, opened 2022-08-21
Created job for sending monthly invoices

maricavor:
Close internetee/registrar_center2#38

### DIFF
--- a/app/jobs/delete_monthly_invoices_job.rb
+++ b/app/jobs/delete_monthly_invoices_job.rb
@@ -1,0 +1,10 @@
+class DeleteMonthlyInvoicesJob < ApplicationJob
+  queue_as :default
+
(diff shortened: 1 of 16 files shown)


vohmar (changes requested):
* in staging set test_registrar setting to true for most of the registrars to avoid flood of montly invoices
* run monthly invoice job in registry: bin/rails r -e staging SendMonthlyInvoicesJob.perform_now
* big error, no invoices created: /home/registry/.rbenv/versions/3.0.3/lib/ruby/gems/3.0.0/bundler/gems/e_invoice-312cac173935/lib/e_invoice/generator.rb:172:in `format': can't convert nil into Float (TypeError)...

vohmar (approved):

PR #2445, opened 2022-09-28
Papertrail domain logs

OlegPhenomenon:
…allback to update related domains


maricavor (approved):

app/models/contact.rb
```diff
@@ -19,6 +19,15 @@ class Contact < ApplicationRecord
   has_many :registrant_domains, class_name: 'Domain', foreign_key: 'registrant_id'
   has_many :actions, dependent: :destroy
 
+  after_save :touch_domain
+  def touch_domain
+    return if domains.empty?
+
+    domains.each do |domain|
+      RefreshUpdateAttributeJob.perform_later('Domain', domain.id, updated_at)
```
maricavor, line 27: Use update_all method here instead of background job with iterator. This method constructs a single SQL UPDATE statement and sends it straight to the database. For example:
domains.update_all(updated_at: updated_at)

maricavor (changes requested):

app/models/contact.rb
OlegPhenomenon: No, in this case we need to use a job. Firstly, the data in the logs is not critical to be processed synchronously. Secondly, the registrar can update the technical contact. This technical contact is tied to each domain. Domains can be one hundred thousand. And in this case, the system will "hang" until a log is created for each domain.

maricavor: In this case maybe this plugin can be helpful: [https://github.com/westonganger/paper_trail-association_tracking](https://github.com/westonganger/paper_trail-association_tracking)
It is used to associate the version of the model and the version of the association that are created in the same transaction.

vohmar:
domain update is triggered if changes are done to admin and tech contacts, does not work for registrant changes

PR #2476, opened 2022-11-04
send expire email for multiyear domains

### LINKED ISSUE #2467: ForceDelete email notifications
In case of soft force delete where force delete starts together with expiration date we send an email notification to domain contacts and additionally to domain@domain and info@domain addresses to alert user about expiry and set restrictions for renewal.

This does not seem to be the case for multi-year registrations where force delete start date is before the expiration date. The same email notification must apply in this case as well i.e. similarly to expiration an email notification must be sent to domain contacts and the additional addresses two days after the force delete start date

OlegPhenomenon:
close #2467

### DIFF
--- a/app/interactions/domains/force_delete/notify_multiyears_expiration_domain.rb
+++ b/app/interactions/domains/force_delete/notify_multiyears_expiration_domain.rb
@@ -1,0 +1,46 @@
+module Domains
+  module ForceDelete
+    class NotifyMultiyearsExpirationDomain < Base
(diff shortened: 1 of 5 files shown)


maricavor (approved):

vohmar:
did not receive any emails. Steps taken

1. set a multiyear domain to forceDelete manually in admin - type=soft
2. go to sidekiq and run the scheduled job

no email in mailcatcher nor actual mailbox. Cannot find errors, but perhaps missing email template?

PR #2483, opened 2022-11-23
Added xml epp console required endpoints and sample xml files

maricavor:


app/controllers/repp/v1/registrar/xml_console_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+        def load_schema_by_prefix(pref, xml)
+          case pref
+          when 'epp-ee'
+            insert_prefix_and_version(xml, pref, '1.0')
+          when 'eis'
+            insert_prefix_and_version(xml, pref, '1.0')
+          when 'contact-ee'
+            insert_prefix_and_version(xml, pref, '1.1')
+          else
+            insert_prefix_and_version(xml, pref, '1.2')
```
OlegPhenomenon, line 48: In my opinion, it is better to keep the versions in some constant

app/models/legal_document.rb
```diff
@@ -69,50 +70,57 @@ def self.remove_duplicates
...
-        unless modified.include?(orig_legal.id)
-          File.delete(new_legal.path) if File.exist?(new_legal.path)
-          new_legal.update(path: orig_legal.path)
-          count += 1
-          Rails.logger.info "File #{new_legal.path} has been removed by Domain "\
-                            "#{new_legal.documentable_id}. Document id: #{new_legal.id}"
-        end
+      LegalDocument.where(documentable_type: 'Domain', documentable_id: orig_legal.documentable_id)
+                   .where(checksum: orig_legal.checksum)
+                   .where.not(id: orig_legal.id)
```
OlegPhenomenon, line 85: Why not use checksum in the above query: `where(documentable_type: 'Domain', documentable_id: orig_legal.documentable_id, checksum: orig_legal.checksum)` ?

app/models/legal_document.rb
```diff
@@ -69,50 +70,57 @@ def self.remove_duplicates
...
+                   .where.not(path: orig_legal.path).each do |new_legal|
```
OlegPhenomenon, line 86: Same here, why not put path: `orig_legal.path` in the query above? The problem with `where` is that it won't work as a filter in the ruby concept. When `where` is called, it will always query the database, even with the sample that was returned from the database. Even if you use `includes`.

app/models/legal_document.rb
```diff
@@ -69,50 +70,57 @@ def self.remove_duplicates
...
+        when Hash
+          id['id']
+        when String
+          JSON.parse(id) rescue id.to_i
+        else
+          id
         end
+      end.flatten.compact.uniq
+      LegalDocument.where(documentable_type: 'Contact', documentable_id: contact_ids)
+                   .where(checksum: orig_legal.checksum)
```
OlegPhenomenon, line 113: Here, too, an additional query is used instead of combining them

OlegPhenomenon (changes requested):

PR #2513, opened 2022-12-13
registrar filters for domain verions of event create

### LINKED ISSUE #2510: Domain_versions - filter with registrar gives no results for create action
if one would look for count of all create actions last month the result is there, but as soon as user sets a value to registrar filter the results are empty (no results).

Steps to reproduce:
1. go to admin/domain_versions
2. set filter parameters: actions = create, created at from = 1.10.2022, created at until = 31.10.2022 and press search
3. now from the displayed results choose any registrar and set that value in the filter for regsitrar and press search
4. No results returned

OlegPhenomenon:
close #2510

### DIFF
--- a/app/controllers/admin/domain_versions_controller.rb
+++ b/app/controllers/admin/domain_versions_controller.rb
@@ -37,7 +37,11 @@ module Admin
       end
       where_s += '  AND 1=0' if registrants == []
       if registrars.present?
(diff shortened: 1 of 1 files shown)


vohmar:
getting internal error in staging with search including regsitrar. For example:

image

PR #2525, opened 2023-01-09
Fixed viewing of long domains in admin portal

### LINKED ISSUE #2520: Admin: long domain names do not fit in the view
/admin/domains/domain_id

image

maricavor:
Close #2520

### DIFF
--- a/app/views/admin/domains/edit.html.erb
+++ b/app/views/admin/domains/edit.html.erb
@@ -9,7 +9,7 @@
 <div class="page-header">
     <div class="row">
         <div class="col-sm-8">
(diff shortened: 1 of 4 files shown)


vohmar:
it seems there is an issue here with this branch - cannot register any new domains on staging, session is closed when trying to add contacts to the registration application. Switched to master branch on registry and epp - did not have this issue, deployed this one - the issue reappeared

vohmar:
might not be related to this particular branch after all as now I have the same issue with another branch and switching back to master did not help either. Unfortunately cannot test the branches until we figure out what is the problem

PR #2578, opened 2023-05-17
Refactored and corrected code for domain force delete

### LINKED ISSUE #2561: ForceDelete - conflicting statuses and too much log
On 9th of April forceDelete was unexpectedly removed from a domain with invlaid email address without the change to the address

The preceeding record for the email validation had this info:

"errors": {"mx": "target host(s) not found"}, "success": true,

notice the conflict - error exists but success is true. The last one is the reason why the FD was lifted. This should not be.
further investigation showd that from the DNS or MX check perspective the email address is actually valid as the domain has IPv4 and IPv6 addresses set. But in reality there is no service behind it. Why is there an error and passed validation simultaneously

There are 1508 records in prod with error message and success = true. Out of those 5 have error "email does not match the regular expression" the rest all share the "target host not found" message

The second part of the problem is the logging. The fd lift job is running every hour and creates a record in domain history for all the FD domains

sample of object_changes records:
2023-04-09 03:56:11.384588+03 | {"updated_at": ["2023-04-09T02:56:12.783+03:00", "2023-04-09T03:56:11.384+03:00"]}

in the logs there are much more detailed record for CheckForceDeleteLift every check and again nothing really changes there except the date. The records are created for every FD domain every hour that is more than 3000 records every day. Need to reduce the data logged and stop creating records in the domain changes when nothing is really changed there.

maricavor:
Close #2561 
- Refactored to improve the readability and maintainability
- Corrected CheckForceDeleteLift job to find domains by their FORCE_DELETE status and not status notes

### DIFF
--- a/app/interactions/actions/a_and_aaaa_email_validation.rb
+++ b/app/interactions/actions/a_and_aaaa_email_validation.rb
@@ -16,19 +16,28 @@ module Actions
     end
 
     def resolve_a_and_aaaa_records(dns_servers:, email_domain:, value:)
-      Resolv::DNS.open({ nameserver: dns_servers }) do |dns|
-        dns.timeouts = ENV['a_and_aaaa_validation_timeout'].to_i || 1
-        ress = nil
+      Resolv::DNS.open(nameserver: dns_servers, ndots: 1, search: []) do |dns|
+        dns.timeouts = (ENV['a_and_aaaa_validation_timeout'] || 1).to_i
 
         case value
         when 'A'
-          ress = dns.getresources email_domain, Resolv::DNS::Resource::IN::A
+          resolve_a_records(dns: dns, domain: email_domain)
         when 'AAAA'
-          ress = dns.getresources email_domain, Resolv::DNS::Resource::IN::AAAA
+          resolve_aaaa_records(dns: dns, domain: email_domain)
+        else
+          []
         end
-
-        ress.map(&:address)
       end
+    end
+
+    def resolve_a_records(dns:, domain:)
+      resources = dns.getresources(domain, Resolv::DNS::Resource::IN::A)
+      resources.map(&:address)
+    end
+
+    def resolve_aaaa_records(dns:, domain:)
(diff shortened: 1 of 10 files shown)


app/interactions/actions/a_and_aaaa_email_validation.rb
```diff
@@ -16,19 +16,28 @@ def check_for_records_value(email:, value:)
...
+          resolve_aaaa_records(dns: dns, domain: email_domain)
+        else
+          []
         end
-
-        ress.map(&:address)
       end
     end
+
+    def resolve_a_records(dns:, domain:)
```
OlegPhenomenon, line 33: domain is not clear, rename to hostname, because Resolv::DNS::Resource::IN::A and Resolv::DNS::Resource::IN::AAAA in context of validation check hostname of email, but not domain itself

app/interactions/actions/email_check.rb
```diff
@@ -22,82 +24,69 @@ def check_email(parsed_email)
     end
 
     def calculate_check_level
-      Rails.env.test? && check_level == 'smtp' ? :mx : check_level.to_sym
+      check_level_sym = check_level.to_sym
```
OlegPhenomenon, line 27: doesn't need here, you can just return check_level.to_sym without extra variable

app/interactions/domains/force_delete_lift/base.rb
```diff
@@ -16,25 +16,26 @@ def lift_force_delete(domain)
       end
 
       def force_delete_condition(domain)
-        domain.force_delete_scheduled? &&
+        force_delete_scheduled?(domain) &&
```
OlegPhenomenon, line 19: what are difference?

OlegPhenomenon (changes requested):

app/interactions/domains/force_delete_lift/base.rb
maricavor: no difference, just better readability

OlegPhenomenon (approved):

PR #2601, opened 2023-07-10
implement validator for company existion

OlegPhenomenon:


maricavor (approved):

vohmar:
the output list of invalid org contacts currently includes all Estonian org type objects no matter the role. But as we set Force Delete only on domains where such an object is in the role of a registrant we need to generate a sub-list or add a role indicator to the output so it would be possible to filter out only the ones important in the context of ForceDelete.

vohmar:
latest test resulted in again with multiple instances of the same entity, but more importantly each entity was matched with only one domain so if a company had 3 domains force delete was set only on one of them

PR #2676, opened 2024-07-17
added endpoint for business registry api

OlegPhenomenon:


vohmar:
* Get available domain name variants - fails to check for variant availability . Lists all the possible variants
* Reserve a domain - fails to check for name availability and successfully creates a token for a registered domainname
* Release reserved domain - returns "error": "Domain not found in reserved list" - status request at the same time returns unpaid (same name, same token (auth type bearer token))
* get registration code for reserved domains - returns ""error": "Domain not found in reserved list"" - because of unpaid status?

OlegPhenomenon:
put here from our discussion:
```
but the problem is that the link is only available in the reservation response, I cannot get it afterwards ie with checking the reservation status. So if i did not save the response there is no way of recovering it
```

vohmar:
* Reserve a domain endpoint returns error 500 for domains available for reservation. It responds correctly with "Domain is not available" for those that are already taken
* Release reserved domains - could not test because reservation does not work (previous bullet point). Same goes for getting the registration code

---

* Get available domain name variants is OK now.  But perhaps we should not announce availability in all zones - ie med.ee is meant strictly for medical organisations

vohmar:
* could not get multiple domains reserved in one request
* status check and registration code request are doubling one another we could perhaps get away with just status check that will return the registration code when the reservation is set on a domain(s)
  * the domain name is missing from the registration_code response: "name": "name",

---

* release reserved domain - OK
* get registration code - OK

PR #2723, opened 2025-01-03
feat: restore company validation for Estonian organizations

OlegPhenomenon:
- Re-enable company validation logic for Estonian organization contacts
- Update company_register gem to use master branch
- Restore and update company validation tests
- Add additional check for Estonian country code in validation

This change ensures that only registered or liquidated Estonian companies can be used when creating organizational contacts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,8 +68,7 @@ gem 'pdfkit'
 gem 'redis'
 gem 'sidekiq', '~> 7.0'
 
-gem 'company_register', github: 'internetee/company_register',
-                        branch: '4-check-for-company-existence'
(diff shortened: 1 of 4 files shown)


vohmar:
Please fix a message that is logged when a company could not be found in the business registry:
for example:
 Force delete set on domain näiteks.ee. Outzone date: 2025-02-13 00:00:00 +0200. Purge date: 2025-03-15 00:00:00 +0200. Invalid ident 1234567. Contact has status No information

The message "Contact has status No information" is hard to understand. Change the message to something like
"Contact not found" or "Contact not found in EE business registry"

maricavor (approved):

Issue #2701, opened 2024-10-24
1.2.25 regulations: optional technical contact


vohmar:
Make technical contact optional over epp and rest-epp.

Currently all .ee domain regulations require specifying technical contact.
Change technical contact to optional on domain create and domain update

the feature needs to be configurable in admin settings to enable deploying it prior to new regulations taking effect

vohmar:
fixed by: https://github.com/internetee/registrar_center2/pull/135

Issue #2702, opened 2024-10-24
1.2.25 regulations: optional admin contact for private registration


vohmar:
Make admin contact optional for private registrations over epp and rest-epp.

Currently .ee domain regulations require specifying admin contact for all registrations
Change admin contact to optional on domain create and domain update for registrations with private registrants. Admin contact is still mandatory for all registrations with org type registrant

the feature needs to be configurable in admin settings to enable deploying it prior to new regulations taking effect

vohmar:
Need to change the settings logic:
we have settings:
Admin contacts min count:
Admin contacts max count:

to achieve needed configurable result with the new conditions:
* admin contact is optional
** except when registrant is an organisation (ident type = org)
** except when registrant is private, but underaged
* admin contact cannot be an organisation (ident type = priv or birthday)


* admin contact min count: 0 should give the general optionality for the admin contacts
** to enforce min count 1 for company registrants add checkbox "Require admin for businesses"
** to enforce min count 1 for minor registrants add checkbox "Require admin for minors" (age is calculated from birthday and if possible from identity code)
* to allow only private admin contacts add setting "allowed admin ident types" with multiple selection list option (in this case priv and birthday should be selected)

vohmar:
fixed by https://github.com/internetee/registry/pull/2748

PR #2730, opened 2025-01-07
Make admin contacts optional for private registrants

OlegPhenomenon:
Close:
#2701 
#2702 

Binded to: https://github.com/internetee/registrar_center2/pull/135

This change makes admin contacts optional for private registrants while keeping them mandatory for organizations. The changes include:

- Updated Domain model validations to make admin and tech contacts optional (min=0) for private registrants
- Added validation rules methods to handle different requirements based on registrant type
- Modified EPP domain creation to support domains without admin contacts for private registrants
- Updated attach_default_contacts to skip adding contacts for private registrants
- Added comprehensive test coverage for:
  - Domain model validations with private/org registrants
  - EPP domain creation without admin contacts for private registrants
  - REPP API contact management for private registrants

This implements the requirement to make admin contacts optional for private registrations of .ee domains while maintaining the existing validation rules for organizations.

### DIFF
--- a/app/models/domain.rb
+++ b/app/models/domain.rb
@@ -164,15 +164,35 @@ class Domain < ApplicationRecord
     max: -> { Setting.dnskeys_max_count },
   }
 
-  validates :admin_domain_contacts, object_count: {
-    min: -> { Setting.admin_contacts_min_count },
-    max: -> { Setting.admin_contacts_max_count },
-  }
+  def self.admin_contacts_validation_rules(for_org:)
+    {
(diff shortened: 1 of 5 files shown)


maricavor (approved):

vohmar:
Current solution lets any domain to be registered without setting admin contacts if the 0 admin contact condition is enabled in admin.

.ee policy requires the admin contact if the registrant is an organisation. It is also needed if the private registrant is a minor (under 18 years old)

how can i enforce the correct admin contact policy?

vohmar:
i was able to set an org type of contact in the role of the admin contact. admin contact can only be private

PR #2748, opened 2025-02-03
feat: add admin contact ident type validation

OlegPhenomenon:
- Add new setting for allowed admin contact ident types
- Add validation for admin contact ident types on domain create/update
- Add UI controls for managing allowed ident types
- Add tests for new validation rules
- Update domain model to respect new settings

The changes allow configuring which identification types (private person, organization, birthday) are allowed for administrative contacts. This is enforced when creating new domains or adding new admin contacts.

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -27,12 +27,25 @@ module Admin
 
     def casted_settings
       settings = {}
-
+    
       params[:settings].each do |k, v|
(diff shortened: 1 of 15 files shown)


vohmar:
fix error message: "Admin contacts are required" replace with "Admin contact is required"

noticed two different error messages for the same error: on domain create:
"Administrative contact with identification type 'org' is not allowed"
on domain update:
"Admin contact can be private person only"

PR #2774, opened 2025-03-31
feat: Add lifted force delete domains to daily admin notification

### LINKED ISSUE #2772: Daily report of new ForceDelete domains
We need to generate and send to all admin users daily report of new domains under ForceDelete process set by the business registry validation process

In the report we need to show

Domainname, FD type (soft/hard), FD reason, ForceDelete date

The report should include FD domains set in the previous day
The report should also include all the domains the ForceDelete was lifted from in the previous day
The report should be sent via email.

OlegPhenomenon:
Closes #2772 

- Add tracking of lifted force delete domains with reason and date in json_statuses_history
- Modify ForceDeleteDailyAdminNotifierJob to include both force deleted and lifted domains
- Update admin mailer template to show separate tables for force deleted and lifted domains
- Update tests to reflect new functionality and fix timing issues with yesterday's data

Key changes:
- Store lift reason and date when canceling force delete
- Add new query method for finding lifted force delete domains
- Split email template into two sections
- Fix tests to properly handle the yesterday time window

### DIFF
--- a/app/interactions/domains/cancel_force_delete/remove_force_delete_statuses.rb
+++ b/app/interactions/domains/cancel_force_delete/remove_force_delete_statuses.rb
@@ -15,6 +15,10 @@ module Domains
         rejected_statuses = domain.statuses.reject { |a| domain_statuses.include? a }
         domain.statuses = rejected_statuses
         domain.skip_whois_record_update = true
+        domain.lift_force_delete_domain_statuses_history_data = {
+          reason: domain.status_notes[DomainStatus::FORCE_DELETE],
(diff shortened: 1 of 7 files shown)


vohmar:
Yesterday FD was set on two domains and 1 FD was removed. This is the email today (next day):
image

the lift is there but neither of the new ones are listed

* some-test-domain-org-23131.ee - soft FD (5.2.2026)
* testdomeen.ee - instant FD (31.3.2025)

The lifted section does not need a table - list of domain names is enough

maricavor (approved):

PR #2808, opened 2025-06-04
Fix stale data in Whois update job by moving callback to after_commit

OlegPhenomenon:
Related #https://github.com/internetee/rest-whois/issues/429

Previously, the UpdateWhoisRecordJob was enqueued from an after_save callback, which caused the job to run before the contact changes were committed to the database. As a result, the WHOIS record was updated with outdated contact data (e.g., previous email address), causing a one-step delay in data synchronization.

This PR moves the enqueue logic to an after_commit callback, ensuring that the job is triggered only after the transaction is fully committed and the latest data is visible to the Sidekiq worker.

Changes made:
- Moved update_related_whois_records from after_save to after_commit.
- Replaced saved_changes with previous_changes for accuracy post-commit.
- Added guard to skip WHOIS update if only ignored fields changed.

This ensures that WHOIS records reflect the actual latest contact information immediately after each update.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -539,12 +539,16 @@ class Contact < ApplicationRecord
   end
 
   def update_related_whois_records
+    Rails.logger.info "\n\n============ update_related_whois_records ==========="
     # not doing anything if no real changes
     ignored_columns = %w[updated_at created_at statuses status_notes]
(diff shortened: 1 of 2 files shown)


vohmar:
the issue with statuses was not reproduced - apparently an issue with the delay of whois update.

PR #2810, opened 2025-06-09
fix issue with idents with special character

mmeest:


mmeest:
Test need to be added.

PR #2811, opened 2025-06-17
Added extra test coverage for validate_dnssec_job

mmeest:


mmeest:
Pull request testing fails. Probable cause is timing in test method:
rails test test/integration/admin_area/white_ips_test.rb:43
Locally that test runs. But on Github possible cause 'page reload' takes longer.

mmeest:
Resolved the problem with ENV['nameserver_validation_timeout']

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
   end
+
+  def test_prepare_validator_configures_dnsruby_resolver_with_correct_parameters
+    job = ValidateDnssecJob.new
+    
+    # To store original environment variable
+    original_timeout = ENV['nameserver_validation_timeout']
+    
+    # Seadista kindel väärtus testiks
+    ENV['nameserver_validation_timeout'] = '4'
```
OlegPhenomenon, line 79: Translate comment to English, because registry is Open Source product.
Also, among the English comments, this one stands out.

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
     @domain.reload
     assert_nil @domain.dnskeys.first.validation_datetime
   end
+
+  def test_prepare_validator_configures_dnsruby_resolver_with_correct_parameters
+    job = ValidateDnssecJob.new
+    
+    # To store original environment variable
```
OlegPhenomenon, line 75: This comment doesn't really add any meaning — it just explains what the code is doing, which is already obvious. So there are two options: either delete the comment, or rewrite it to explain what the nameserver_validation_timeout value does and how it affects the test.

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    original_timeout = ENV['nameserver_validation_timeout']
+    
+    # Seadista kindel väärtus testiks
+    ENV['nameserver_validation_timeout'] = '4'
+    
+    resolver = job.send(:prepare_validator, "8.8.8.8")
+    
+    assert_instance_of Dnsruby::Resolver, resolver
```
OlegPhenomenon, line 83: Are we testing type safety or functionality here? Please clarify, because otherwise I don't see the point of this check.

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    domain = Domain.create!(
+      name: "test.test",
+      registrar: registrars(:bestnames),
+      registrant: @domain.registrant,
+      period: 1,
+      period_unit: 'y',
+      valid_to: 1.year.from_now
+    )
+
+    # Add DNSKEY to domain
```
OlegPhenomenon, line 103: remove this comment

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    # Create a StringIO to capture log output
+    log_output = StringIO.new
+    logger = Logger.new(log_output)
+    logger.level = Logger::INFO
+
+    # Create a job instance and set its logger
+    job = ValidateDnssecJob.new
+    job.define_singleton_method(:logger) { logger }
+
+    # Run the job
```
OlegPhenomenon, line 115: remove it

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    job.perform(domain_name: domain.name)
+
+    # Verify that the domain was skipped
+    assert_match /No related nameservers for this domain/, log_output.string
```
OlegPhenomenon, line 119: I think you've overcomplicated the test a bit. If you look at the file app/jobs/validate_dnssec_job.rb, the dns key object has a field called validation_datetime, which gets updated when the validation is successful. You could’ve used that for the check.

Of course, if the goal is to verify the logs, then tracking logs makes sense too. But what's really important here is that the job didn’t make any changes to the database record.

We actually need two clear tests:
One test that confirms the table was updated when a nameserver is present.
And this current test should prove that validation was skipped, meaning no changes happened in the table at all.

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+  end
+
+  def perform_without_domain_name_executes_else_block
+    # Use existing domain fixture
```
OlegPhenomenon, line 123: remove it

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    domain = domains(:shop)
+    
+    # Add DNSKEY if not present
```
OlegPhenomenon, line 126: you can rewrite this part more like "ruby-style":
```
domain.dnskeys << @dnskey unless domain.dnskeys.any?
```

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+      domain.nameservers.create!(
+        hostname: "ns.#{domain.name}",
+        ipv4: ["192.0.2.1"],
+        validation_datetime: Time.zone.now
+      )
+    end
+  
+    ValidateDnssecJob.perform_now()
+    
+    assert true, "Job finished without errors"
```
OlegPhenomenon, line 142: ’m not quite sure I understand the point of this test. What happens if there’s no nameserver — does it still complete without errors? What exactly is the success criteria here? Just checking that the job runs?

What we really need is to confirm that the job actually did its job — meaning the data we expect to change has actually changed (or hasn’t, depending on the case).

I know the job we’re testing includes exception handling, but those should have their own separate tests. Here, the main check should be whether the expected data changes happened — that's the real sign the job worked correctly.

test/jobs/validate_dnssec_job_test.rb
```diff
@@ -68,4 +68,77 @@ def test_job_should_not_set_validation_datetime_if_validation_is_invalid
...
+    # Add nameserver if not present
+    unless domain.nameservers.any?
+      domain.nameservers.create!(
+        hostname: "ns.#{domain.name}",
+        ipv4: ["192.0.2.1"],
+        validation_datetime: Time.zone.now
+      )
+    end
+  
+    ValidateDnssecJob.perform_now()
```
OlegPhenomenon, line 140: Remove the parentheses.

test/jobs/validate_dnssec_job_test.rb
mmeest: Removed the unnecessary comment.

test/jobs/validate_dnssec_job_test.rb
mmeest: While Ruby is dynamically typed language, so it is meant that code behaves correctly at runtime. It is functional guarantee we can use resolver object in next steps.

test/jobs/validate_dnssec_job_test.rb
mmeest: Removed!

mmeest:
> Please read my feedback, if you have some question, you can contact with me

Thank you for really valuable feedback!

OlegPhenomenon (approved):

PR #2825, opened 2025-07-17
feat(admin): notify registrar about domain status changes

### LINKED ISSUE #2823: Send poll message when admin adds or removes statuses on a domain
Currently there is no indication to the registrar about changes with domain statuses when added or removed by admin in admin ui (except for force delete).

Add poll messages for any changes with domain statuses done by admin via the admin.

Poll text: <status> set on domain <domain name> / <status> removed from domain <domain name>

OlegPhenomenon:
close #2823

- Add inform_registrar_about_status_changes method to detect added and removed statuses after domain update.
- Send notification to registrar with details of which statuses were set and which were removed.
- Integrate notification logic into domain update action.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 ruby:3.0.3-slim-buster
+FROM --platform=linux/amd64 ruby:3.0.3-bullseye
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update > /dev/null && apt-get install -y --no-install-recommends > /dev/null \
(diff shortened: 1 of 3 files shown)


vohmar:
How about sending separate poll message about each added and removed status (with the exception of sets of statuses like Force Delete). So now it shoes this: Set on orgrant4.ee: serverRenewProhibited, serverTransferProhibited. But instead there could be 2 separate messages

* serverRenewProhibited set on orgrant4.ee
* serverTransferProhibited set on orgrant4.ee

and same for removing but perhaps use canceled termin for that

* serverRenewProhibited is cancelled on domain orgrant4.ee

this would unify the style with existing Force delete messages 

* Force delete set on domain orgrant4.ee. Outzone date: 2025-11-15 00:00:00 +0200. Purge date: 2025-12-15 00:00:00 +0200
* Force delete is cancelled on domain orgrant4.ee

and hopefully will make it easier to parse for registrars

PR #2829, opened 2025-07-25
Added new tests for admin/domains_controller

mmeest:
Added new tests for admin/domains_controller methods

- index
- update
- versions
- keep

also private methods:

- domain_params
- ignore_empty_statuses
- normalize_search_parameters


test/integration/admin_area/domains_controller_test.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def test_index_renders_successfully
+    get admin_domains_path
+    assert_response :success
+    assert_match @domain.name, response.body
+  end
+
+  def test_index_with_status_filter
+    status = @domain.statuses.first || 'client_hold'
```
OlegPhenomenon, line 21: This isn’t a critical mistake, I used to do the same quite often myself, and to be honest, I still do it sometimes when laziness gets the better of my technical debt. But let me pose a question for reflection: what consequences or complications might arise if the requirements suddenly change and we have to rename the status from 'client_hold' to, say, 'client_run'? :)

Hint: this relates to the DRY (Don't Repeat Yourself) programming principle.

OlegPhenomenon (approved):

test/integration/admin_area/domains_controller_test.rb
mmeest: Instead of hard coded string in test it would be more practical and reusable to define it in fixtures.

PR #2860, opened 2025-09-18
Adding test coverage to sso/tara_controller

mmeest:
New tests added


vohmar:
let's remove the useless controller code  then @maricavor ?

PR #2911, opened 2026-04-01
Accred api OIDC

maricavor:


OlegPhenomenon (approved):

vohmar:
github tests are failing. Copilots suggestions to avoid these regular failures of selenium tests:

Key Recommendations

- Replace XPath with ID/class-based selectors — they're more stable
- Add explicit waits before interacting with elements using assert_selector
- Avoid hardcoded DOM paths — maintain them via data-testid attributes
- Use Capybara's built-in retry logic instead of raw Selenium calls

PR #2918, opened 2026-04-22
Pdf invoice generator for business registry

OlegPhenomenon:


vohmar:
Codex review:

app/models/reserve_domain_invoice/pdf_generator.rb
The PDF total does not match the amount that is actually created in billing. create_invoice_with_domains sends domain_names.count * DEFAULT_AMOUNT as the invoice total, but this PDF generator treats the same row prices as VAT-exclusive and adds Setting.registry_vat_prc on top. For example, with the default VAT setting of 0.2, a one-domain reservation billed at 10.00 will be rendered in the PDF as 12.00. We should either align the billing amount and PDF calculation, or treat DEFAULT_AMOUNT consistently as VAT-inclusive and split VAT out from that total.

app/controllers/api/v1/business_registry/reserved_domains_invoice_pdf_controller.rb
For an official invoice PDF, amount_paid should not come from request params. The caller can currently make the PDF show any paid amount or overpayment, even though invoice_state only verifies that the invoice is paid. This should be derived from billing data or from the stored invoice/payment record so the rendered payment amount cannot differ from the actual transaction.

maricavor (approved):

